PR #9, opened 2019-04-22
Add CompactVersion type

markelliot:


carterkozak:
Is there a ticket describing why we need to pack bits?

markelliot:
On occasion we need to store versions in a data store, and the string form of these versions are not lexicographically orderable, making it hard for us to sort in the data store or create indexes. Several of our internal systems approximate the representation by multiplying out the versions, but this compact form produces something more suitable for purpose. Rather than have every consumer implement this functionality, it seemed wiser to contribute the utility class.

________________________________

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -0,0 +1,211 @@
...
+
+    private static String generateVersionString(int major, int minor, int patch, SlsVersionType type,
+            int rcNumber, int distanceFromVersion) {
+        StringBuilder sb = new StringBuilder();
+        sb.append(major).append(".").append(minor).append(".").append(patch);
+        if (type.isReleaseCandidate()) {
+            sb.append("-rc").append(rcNumber);
+        }
+        if (type.isSnapshot()) {
+            sb.append("-").append(distanceFromVersion).append("-gaaaaaa");
```
uschi2000, line 190: it was probably a mistake when I made `equals()` deviate from string equality. now that we'll have to live with this bug, can you please add a little warning on top of this class: `CompactVersion.from().toSlsVersion()` round-trips all equality and comparable concerns, but does not round-trip string equality.

uschi2000: (on the other hand .. maybe it wasn't a mistake because that deviation is what allows us to define this compact format in the first place)

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -0,0 +1,211 @@
...
+
+    private final long msb;
+    private final long lsb;
+
+    private CompactVersion(long msb, long lsb) {
+        this.msb = msb;
+        this.lsb = lsb;
+    }
+
+    public long getMsb() {
```
uschi2000, line 64: maybe safer (in terms of lexicographic ordering use cases) to expose only the byte[] and not the underlying msb/lsb?

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -0,0 +1,211 @@
...
+            case RELEASE_CANDIDATE_SNAPSHOT:
+                return 0;
+            case NON_ORDERABLE:
+                throw new SafeIllegalArgumentException("Unable to store NON_ORDERABLE types in CompactVersion");
+        }
+        throw new SafeIllegalArgumentException("Unknown SlsVerisonType", SafeArg.of("slsVersionType", type));
+    }
+
+    private static long encode20b(int value, @CompileTimeConstant String component) {
+        Preconditions.checkArgument(value >= 0 && value < 1_048_576,
```
uschi2000, line 116: add test for this, and also write in the class-level docs

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -0,0 +1,211 @@
...
+ * <code>
+ * LSB SafeLong:
+ * 20 bits: distance from release
+ *  2 bits: priority1 (1 = RELEASE_CANDIDATE_SNAPSHOT, 0 = else)
+ * 20 bits: RC number
+ *  2 bits: priority2 (2 = RELEASE_SNAPSHOT, 1 = RELEASE, 0 = RELEASE_CANDIDATE)
+ *  8 bits: lowest 8 bits of patch
+ *
+ * MSB SafeLong:
+ * 12 bits: highest 8 bits of patch
```
uschi2000: highest `12` bits of patch

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -0,0 +1,211 @@
...
+            rcNumber = 0;
+            distanceFromVersion = version.firstSequenceVersionNumber().orElse(0);
+        }
+
+        long lsb = encode20b(distanceFromVersion, "distanceFromVersion")
+                + (priority1(version.getType()) << 20)
+                + (encode20b(rcNumber, "rcNumber") << 22)
+                + (priority2(version.getType()) << 42)
+                + ((patch & 0xFF) << 44);
+        long msb = ((patch & 0xFFF00) >> 8)
```
uschi2000, line 89: Haven't verified 100%, but have you convinced yourself that you won't run into ordering problems when a leading `1` bit of the msb or lsb long indicates a 2-complement negative number? (this would be fine for lexicographic comparison, but would screw up the current implementation of `compareTo` because it's in terms of longs.)

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -0,0 +1,211 @@
...
+ * {@link OrderableSlsVersion}. To enable representing bytes as two SafeLong values, this implementation packs at most
+ * 53 bits into a single {@code long}.
+ *
+ * <p>Bits are allocated as follows, from lowest bits to highest:
+ * <code>
+ * LSB SafeLong:
+ * 20 bits: distance from release
+ *  2 bits: priority1 (1 = RELEASE_CANDIDATE_SNAPSHOT, 0 = else)
+ * 20 bits: RC number
+ *  2 bits: priority2 (2 = RELEASE_SNAPSHOT, 1 = RELEASE, 0 = RELEASE_CANDIDATE)
```
uschi2000: 3=?

uschi2000 (commented):
few questions for you

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: Yes, the tests use the maximum values and ensure round-tripping

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: 3 is an unused value

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: another option is to skip including these methods for now (I only need one direction to work for my use-case)

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: I think it'll be pretty standard to represent these as two longs in data stores (and in the context of AtlasDB as two longs used in a compound key), and so we can skip some plumbing by using longs.

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: done

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: noted unused in docs, and same for unused `priority1` values.

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: commented in the javadoc on this method, but lmk if you want me to remove the toSlsVersion stuff entirely

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: done

markelliot:
@uschi2000 take another look, please

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
uschi2000: this works "accidentally" because you only use the top 50 and 52 bits. if the 0th bit (the highest-order one) were ever `1`, then you get a negative long. maybe add to class-level docs, `Note that the correctness of this implementation relies on the fact that strictly less than 64 bits are used for the encoding, see {@link #compareTo}` and this to compareTo, `Note that the correctness of the comparison implementation relies on the fact that the msb and lsb never have a leading 1 because we only use strictly less than 64 bits for the encoding. Without this constraint, a leading 1 in the (two-complement) msb or lsb would would represent a negative number and thus render this compareTo implementation incorrect.`

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
uschi2000: ok

uschi2000 (approved):
one more comment please, otherwise +1

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
markelliot: chose slightly different text, but added a class note.

also noting that there are tests that ensure the value does not exceed 53 bits in the maximum case, and that values that would require more than 20 bits of storage throw, the combination of which should test this implementation detail.

uschi2000: yeah, agree the status quo is well covered, but I think in this kind of scenario it's important that we document *why* we have this constraint. otherwise, in 2 years someone may come along, add a few bits, and change the test to 64 bits.

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -0,0 +1,222 @@
...
+ *  2 bits: priority2 (2 = RELEASE_SNAPSHOT, 1 = RELEASE, 0 = RELEASE_CANDIDATE; 3 is an unused value)
+ *  8 bits: lowest 8 bits of patch
+ *
+ * MSB SafeLong:
+ * 12 bits: highest 12 bits of patch
+ * 20 bits: minor
+ * 20 bits: major
+ * </code>
+ *
+ * <p><b>Note</b>: the correctness of the implementation of {@link #compareTo(CompactVersion)} depends on the
```
uschi2000, line 46: perfect, much more eloquent than my version, thanks

PR #76, opened 2019-09-09
Change ImmutablesStyle package

pkoenig10:
This conflicts with an internal `ImmutablesStyle` in the same package.


iamdanfox (approved):
This seems like a better practise - only declaring classes within the packages we control in this repo.

PR #92, opened 2019-11-01
Fix very nonobvious `check` behavior for `OrderableSlsVersion` and `NonOrderableSlsVersion`

lycarter:
## Before this PR

`check` is not overridden by `OrderableSlsVersion` or `NonOrderableSlsVersion`, which means it's the same as calling `SlsVersion.check` which has different expected behavior in a confusing and unexpected way.

Additionally, `SlsVersion::check` caught an `IllegalArgumentException` when what was actually being thrown was a `SafeIllegalArgumentException` in some cases.

I also cleaned up a couple of things the compiler was complaining about

## After this PR
==COMMIT_MSG==
Implement `check` method for `OrderableSlsVersion` and `NonOrderableSlsVersion` with correct semantics. Also fixes a case where this method would throw a `SafeIllegalArgumentException` unexpectedly.
==COMMIT_MSG==

### DIFF
--- a/sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
@@ -112,7 +112,7 @@ public final class CompactVersion implements Comparable<CompactVersion> {
         throw new SafeIllegalArgumentException("Unknown SlsVersionType", SafeArg.of("slsVersionType", type));
     }
 
-    private static long encode20b(int value, @CompileTimeConstant String component) {
+    private static long encode20b(int value, @CompileTimeConstant final String component) {
         Preconditions.checkArgument(value >= 0 && value < 1_048_576,
                 "version component must be positive and not exceed 20 bits of value",
                 SafeArg.of(component, value));
@@ -151,8 +151,8 @@ public final class CompactVersion implements Comparable<CompactVersion> {
                 firstSeq = OptionalInt.of(rcNumber);
                 secondSeq = OptionalInt.of(distanceFromVersion);
                 break;
-            default:
-                // nothing
+            case RELEASE:
+            case NON_ORDERABLE:
                 break;
         }
         return new OrderableSlsVersion.Builder()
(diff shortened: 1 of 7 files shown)


sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
```diff
@@ -43,7 +46,7 @@ public static boolean check(String coordinate) {
         try {
             valueOf(coordinate);
             return true;
-        } catch (IllegalArgumentException e) {
+        } catch (SafeIllegalArgumentException e) {
```
iamdanfox, line 49: I think I'd leave this as IAE just to be on the safe - they might come from inside immutables too?

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -112,7 +112,7 @@ private static long encodePriority2(SlsVersionType type) {
         throw new SafeIllegalArgumentException("Unknown SlsVersionType", SafeArg.of("slsVersionType", type));
     }
 
-    private static long encode20b(int value, @CompileTimeConstant String component) {
+    private static long encode20b(int value, @CompileTimeConstant final String component) {
```
iamdanfox: Would prefer to omit the `final` here if possible?

iamdanfox (approved):
Looks reasonable to me!

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
lycarter: everything here depends on logsafe, and I ran into this error *because* deep down something was throwing a SIAE

It also passes all of the unit tests - if you can point to a codepath where it wouldn't work, I'm more than happy to adjust, but I think this fixes everything

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
lycarter: done - I hadn't read carefully, thought this was a compiler error but it was only a warn 👍

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
iamdanfox: Fair enough - immutables would throw NullPointerExceptions anwyay, so this doesn't really buy us anything.

iamdanfox (approved):

lycarter:
thanks @iamdanfox !

PR #146, opened 2020-01-20
remove jackson dep, only pull in annotations

ferozco:
## Before this PR
We pulled in all jackson-databind so that we could get the annotations

## After this PR
Only pull in the required annotations

## Possible downsides?
N/A


carterkozak:
:+1:

carterkozak:
dozer?

PR #311, opened 2020-08-14
Don't throw exceptions in safeValueOf

pkoenig10:
## Before this PR
`OrderableSlsVersions#safeValueOf` can throw an exception if the provided value is `null`.

This caused some difficulty debugging internally because this method threw a NPE before the callers could gracefully handle the absent return value.

## After this PR
`OrderableSlsVersions#safeValueOf` does not throw exceptions and always returns `Optional.empty()` for invalid input.

Callers should be able to assume that a method named `safeValueOf` returning an `Optional` will not throw.


fawind (approved):
Sounds reasonable. Thanks for following up!

fawind:
Closed and reopened to make the cla bot happy

PR #400, opened 2021-01-26
OrderableSlsVersion implements Comparable

pkoenig10:
## Before this PR
It is cumbersome to use `OrderableSlsVersion` with `Comparator` combinators because `OrderableSlsVersion` is not `Comparable`, even though it has a natural and canonical `Comparator` implementation.

## After this PR
`OrderableSlsVersion` implements `Comparable`.

### DIFF
--- a/changelog/@unreleased/pr-400.v2.yml
+++ b/changelog/@unreleased/pr-400.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: '`OrderableSlsVersion` now implements `Comparable<OrderableSlsVersion>`.'
+  links:
+  - https://github.com/palantir/sls-version-java/pull/400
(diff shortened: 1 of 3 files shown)


fawind (approved):
Looks like a good addition to me!

Do you know why we didn't implement this in the first place but used the `VersionComparator.INSTANCE` instead?

pkoenig10:
Traced this back to https://g.p.b/foundry/apollo/pull/42.

Looks like `OrderableSlsVersion` initially implemented the raw `Comparable` class with a comment that said:
```
// TODO(rfink) Need to use Comparable<T> with generics
```

So I can't imagine there's a good reason why we didn't just implement `Comparable<OrderableSlsVersion>`.

Maybe @uschi2000 can comment.

PR #455, opened 2021-06-07
SlsVersionMatcher deserialization no longer obliterates our GC (aka eliminate one NumberFormatException)

iamdanfox:
## Before this PR

For the last couple of weeks, the Apollo service I work on (AS) has been _obliterated_ by stop the world GCs.  We know exactly which PR made the change, but have struggled to identify _why_ this PR caused so much GC trouble, because it didn't introduce larger volume of network requests, and response sizes are pretty comparable. This is painful enough that it is triggering monitors, auto-filed tickets (e.g. PDS-186361) and is one of our highest-priority stability things to fix.

I took a Java Flight Recording (JFR) using WC's endpoint, and opened it in Zulu Mission Control, to see the following warning.  ZMC is telling us that 2.47 GB of our heap comes from Object arrays, and these are being created by the `fillInStackTraceMethod`, invoked from the `SlsVersionMatcher.safeValueOf` method.

image (1)

This is the smoking gun, because it turns out that the PR did involve deserializing a _lot_ of SlsVersionMatcher objects (i.e. 1.x.x, 2.x.x).  Turns out, when you deserialize a string into a SlsVersionMatcher, we hit this spicy method:

```java
private static OptionalInt parseInt(String maybeInt) {
    try {
        return OptionalInt.of(Integer.parseInt(maybeInt));
    } catch (NumberFormatException e) { // 🌶🌶🌶 <- exceptions as control flow!!!!
        return OptionalInt.empty();
    }
}
```

This is particularly heartbreaking because the string `maybeInt` is usually either an int (single-digit or multi-digit), or the literal `x`.  It seems like it's invocations of `Integer.parseInt("x")` over and over again which are spewing these NumberFormatExceptions.

## After this PR
==COMMIT_MSG==
`SlsVersionMatcher#safeValueOf` now allocates 94% less memory per invocation, by avoiding throwing and catching a `NumberFormatException`. Specifically, it would allocate ~2192 bytes for every single invocation in my benchmark, and now it allocates 152 bytes per op. It's also ~10x faster.
==COMMIT_MSG==

Correctness is verified by using the property-based testing library jQwik (which @CRogers introduced me to) to fuzz my 'hand-rolled' parser against the regex parser and ensure they're the same.

Performance and GC numbers are provided by JMH benchmarks which can be run locally.  I ran it through a YourKit profiler just for fun (which revealed the majority of time was being spent in regex-land, which is why I went for the hand rolled impl).

## Possible downsides?
- hand-rolled parser code is harder to read than the naive regex stuff.... however I think the jqik tests give us _tons_ of confidence of the correctness, because it literally exercises all kinds of pernicious edge cases and guarantees equivalence


benchmarks/src/jmh/java/com/palantir/sls/versions/SlsVersionMatcherBenchmark.java
```diff
@@ -0,0 +1,93 @@
...
+
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Warmup(iterations = 3, time = 3, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 3, time = 3, timeUnit = TimeUnit.SECONDS)
+@Fork(1)
+@Threads(4)
+@SuppressWarnings({"checkstyle:hideutilityclassconstructor", "checkstyle:VisibilityModifier"})
+public class SlsVersionMatcherBenchmark {
```
iamdanfox, line 48: ## DEVELOP

```
Benchmark                                                                             Mode  Cnt     Score      Error   Units
SlsVersionMatcherBenchmark.matcher_with_all_numbers                                   avgt    3   237.135 ±  190.674   ns/op
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.alloc.rate                    avgt    3  6841.956 ± 5528.811  MB/sec
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.alloc.rate.norm               avgt    3   496.028 ±    0.009    B/op
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Eden_Space           avgt    3  6912.865 ± 5819.395  MB/sec
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Eden_Space.norm      avgt    3   501.141 ±   20.436    B/op
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Survivor_Space       avgt    3     0.025 ±    0.076  MB/sec
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Survivor_Space.norm  avgt    3     0.002 ±    0.004    B/op
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.count                         avgt    3   164.000             counts
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.time                          avgt    3    93.000                 ms
SlsVersionMatcherBenchmark.matcher_with_one_x_x                                       avgt    3  3579.245 ± 2190.375   ns/op
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.alloc.rate                        avgt    3  2002.960 ± 1233.671  MB/sec
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.alloc.rate.norm                   avgt    3  2192.153 ±    0.043    B/op
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Eden_Space               avgt    3  2029.565 ± 1119.353  MB/sec
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Eden_Space.norm          avgt    3  2221.447 ±  154.314    B/op
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Survivor_Space           avgt    3     0.009 ±    0.050  MB/sec
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Survivor_Space.norm      avgt    3     0.009 ±    0.048    B/op
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.count                             avgt    3    58.000             counts
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.time                              avgt    3    33.000                 ms
```


## LAST COMMITS

```
Benchmark                                                                                  Mode  Cnt      Score      Error   Units
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers                                   avgt    3     22.810 ±   11.652   ns/op
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.alloc.rate                    avgt    3   8030.156 ± 4045.447  MB/sec
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.alloc.rate.norm               avgt    3     56.002 ±    0.001    B/op
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.churn.G1_Eden_Space           avgt    3   8051.053 ± 5019.261  MB/sec
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.churn.G1_Eden_Space.norm      avgt    3     56.146 ±   17.280    B/op
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.churn.G1_Survivor_Space       avgt    3      0.018 ±    0.091  MB/sec
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.churn.G1_Survivor_Space.norm  avgt    3     ≈ 10⁻⁴               B/op
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.count                         avgt    3    134.000             counts
SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.time                          avgt    3     88.000                 ms
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x                                       avgt    3     41.704 ±    6.885   ns/op
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.alloc.rate                        avgt    3  11908.726 ± 1854.344  MB/sec
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.alloc.rate.norm                   avgt    3    152.005 ±    0.007    B/op
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.churn.G1_Eden_Space               avgt    3  11929.484 ± 1544.694  MB/sec
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.churn.G1_Eden_Space.norm          avgt    3    152.271 ±    8.645    B/op
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.churn.G1_Survivor_Space           avgt    3      0.032 ±    0.035  MB/sec
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.churn.G1_Survivor_Space.norm      avgt    3     ≈ 10⁻³               B/op
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.count                             avgt    3    214.000             counts
SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.time                              avgt    3    143.000                 ms
SlsVersionMatcherBenchmark.matcher_with_all_numbers                                        avgt    3    274.983 ±   69.234   ns/op
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.alloc.rate                         avgt    3   5895.971 ± 1487.802  MB/sec
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.alloc.rate.norm                    avgt    3    496.023 ±    0.015    B/op
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Eden_Space                avgt    3   5959.603 ± 2776.263  MB/sec
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Eden_Space.norm           avgt    3    501.325 ±  111.866    B/op
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Survivor_Space            avgt    3      0.015 ±    0.097  MB/sec
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Survivor_Space.norm       avgt    3      0.001 ±    0.008    B/op
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.count                              avgt    3    117.000             counts
SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.time                               avgt    3     77.000                 ms
SlsVersionMatcherBenchmark.matcher_with_one_x_x                                            avgt    3    262.229 ±  192.393   ns/op
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.alloc.rate                             avgt    3   5588.739 ± 4057.361  MB/sec
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.alloc.rate.norm                        avgt    3    448.021 ±    0.011    B/op
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Eden_Space                    avgt    3   5605.757 ± 4368.932  MB/sec
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Eden_Space.norm               avgt    3    449.350 ±   24.536    B/op
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Survivor_Space                avgt    3      0.017 ±    0.038  MB/sec
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Survivor_Space.norm           avgt    3      0.001 ±    0.003    B/op
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.count                                  avgt    3    110.000             counts
SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.time                                   avgt    3     72.000                 ms
```

benchmarks/src/jmh/java/com/palantir/sls/versions/SlsVersionMatcherBenchmark.java
```diff
@@ -0,0 +1,93 @@
...
+    @Setup
+    public final void before() {}
+
+    @TearDown
+    public final void after() {}
+
+    @Benchmark
+    public static void matcher_with_one_x_x(Blackhole blackhole) {
+        Optional<SlsVersionMatcher> answer = RegexMatcherParser.safeValueOf("1.x.x");
+        blackhole.consume(answer);
```
carterkozak:
```suggestion
    public static Optional<SlsVersionMatcher> matcher_with_one_x_x(Blackhole blackhole) {
        return RegexMatcherParser.safeValueOf("1.x.x");
```

These benchmarks can be simplified, both are equivalent.

benchmarks/src/jmh/java/com/palantir/sls/versions/SlsVersionMatcherBenchmark.java
```diff
@@ -0,0 +1,93 @@
...
+@Fork(1)
+@Threads(4)
+@SuppressWarnings({"checkstyle:hideutilityclassconstructor", "checkstyle:VisibilityModifier"})
+public class SlsVersionMatcherBenchmark {
+
+    @Setup
+    public final void before() {}
+
+    @TearDown
+    public final void after() {}
```
carterkozak:
```suggestion
```

sls-versions/src/main/java/com/palantir/sls/versions/HandRolledMatcherParser.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.OptionalInt;
+
+/**
+ * A hand-rolled implementation of {@link RegexMatcherParser}.
+ *
+ * Uses two-element int[] arrays as the 'result' type to represent a successful/failed parse, where result[0] always
+ * contains the index into the string we're looking at, and result[1] contains the value returned by the parser
+ * combinator.
+ */
+final class HandRolledMatcherParser {
```
iamdanfox: I'm open to the argument that this is overkill... 

I kinda just wanted to write it to see how good we could make this, given that this class is so heavily used inside Apollo. The use of jqwik is really what tips this over into being viable imo - I've found that changing _any_ line in this file will trip tests... I'm pretty confident this cannot diverge from the reference impl

carterkozak: Haven't reviewed the logic yet, but I'd rename this to `SlsVersionMatcherParser`. The previous implementation was hand-rolled to delegate to the java regex implementation with a specific pattern ;-)

sls-versions/src/main/java/com/palantir/sls/versions/HandRolledMatcherParser.java
```diff
@@ -0,0 +1,148 @@
...
+
+    public static Optional<SlsVersionMatcher> safeValueOf(String string) {
+        OptionalInt major = OptionalInt.empty();
+        OptionalInt minor = OptionalInt.empty();
+        OptionalInt patch = OptionalInt.empty();
+
+        int[] result = new int[] {0};
+
+        // major
+        result = numberOrX(string, result[0]);
```
carterkozak: Dead array allocation. I'm not sure why we reuse the same reference for subsequent arrays, it might. be easier to follow if we used `int[] major =...`. That said, I'm not sure why we're using arrays when `long` is a great `int[2]`.
```suggestion
        // major
        int[] result = numberOrX(string, 0);
```

sls-versions/src/main/java/com/palantir/sls/versions/RegexMatcherParser.java
```diff
@@ -0,0 +1,57 @@
...
+package com.palantir.sls.versions;
+
+import static com.palantir.logsafe.Preconditions.checkNotNull;
+
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+final class RegexMatcherParser {
```
carterkozak: Should we move this to test sources?

iamdanfox: Done! Had to collapse the :benchmarks project into :sls-versions because it could no longer access this class once it was moved into :sls-versions' test sourceset, but I think this is fine because jmh stuff is separated into a nice jmh source set.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
```diff
@@ -70,38 +71,32 @@ public static SlsVersionMatcher valueOf(String value) {
...
+                    "Not a valid matcher, a patch version is specified, yet a major or minor is not specified",
+                    SafeArg.of("matcher", maybeMatcher));
+            return Optional.empty();
+        }
+        if (maybeMatcher.getMinorVersionNumber().isPresent()
+                && !maybeMatcher.getMajorVersionNumber().isPresent()) {
+            // String contains a pattern where major version is underspecified. Example: x.2.x
+            log.info(
+                    "Not a valid matcher, a minor version is specified, yet a major is not specified",
+                    SafeArg.of("matcher", maybeMatcher));
```
iamdanfox, line 96: all the logic here is unchanged, it just got pulled out to a `maybeCreate` method so that it's usable from the Regex & new hand-rolled impls.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
```diff
@@ -70,38 +71,32 @@ public static SlsVersionMatcher valueOf(String value) {
...
     public static Optional<SlsVersionMatcher> safeValueOf(String value) {
         checkNotNull(value, "value cannot be null");
+        return SlsVersionMatcherParser.safeValueOf(value);
+    }
 
-        Matcher matcher = PATTERN.matcher(value);
-        if (!matcher.matches()) {
+    static Optional<SlsVersionMatcher> maybeCreate(
+            String value, OptionalInt major, OptionalInt minor, OptionalInt patch) {
+        SlsVersionMatcher maybeMatcher = ImmutableSlsVersionMatcher.of(value, major, minor, patch);
```
iamdanfox, line 79: this line did change though - it used to be a builder and now it's a @Value.Parameter single-shot construct-it-all invocation.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
```diff
@@ -164,14 +159,6 @@ public int compare(OrderableSlsVersion version) {
         return OrderableSlsVersion.safeValueOf(getValue());
     }
 
-    private static OptionalInt parseInt(String maybeInt) {
-        try {
-            return OptionalInt.of(Integer.parseInt(maybeInt));
-        } catch (NumberFormatException e) {
-            return OptionalInt.empty();
```
schlosna: I haven't read through the full PR, but I'm curious if you looked at Guava's [`Ints.tryParse(String)`](https://guava.dev/releases/19.0/api/docs/com/google/common/primitives/Ints.html#tryParse(java.lang.String)) which returns null instead of using expensive exceptions for control flow.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcherParser.java
```diff
@@ -0,0 +1,146 @@
...
+            result[1] = MAGIC_X_NUMBER;
+            return result;
+        }
+
+        result = number(string, startIndex);
+        if (result[1] != Integer.MIN_VALUE) {
+            return result;
+        }
+
+        return new int[] {startIndex, Integer.MIN_VALUE};
```
CRogers: I think if you use a type instead of an array it might be easier to read and also less memory, as arrays are objects in java but also contain an extra length field that wouldn't be needed with a type.

CRogers: Ah, apparently HotSpot treats arrays specially and stores the length in object header, so I don't think it would be more memory efficient.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
iamdanfox: Yep I think I would be happy to get that as a first-step, as the GC pressure is pretty savage at the moment. Only concern is that it would pull in a guava dep.  Here's a PR up anyway in case maintainers prefer it https://github.com/palantir/sls-version-java/pull/456

carterkozak: I've approved the other PR, happy to merge that first if you like. Up to you if you'd like to further improve things with this PR -- perhaps worth benchmarking the difference?

iamdanfox:
Have merged in develop, so we're now benchmarking against the guava method. The guava thing seems to always be allocating ~424 bytes per op, whereas the hand rolled one is either 56 or 200 bytes per op. (and obviously theres a speedup too, but I don't really care about that).

It's a little bit marginal, but I _think_ this might be worth it tbh. I can pick up the new release and try it in Apollo to know for sure.

```diff
 Benchmark                                                                                  Mode  Cnt      Score       Error   Units
 SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers                                   avgt    3     23.481 ±     4.638   ns/op
 SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.alloc.rate                    avgt    3   7792.647 ±  1544.603  MB/sec
+SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.alloc.rate.norm               avgt    3     56.003 ±     0.001    B/op
 SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.churn.G1_Eden_Space           avgt    3   7834.409 ±  1608.531  MB/sec
 SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.churn.G1_Eden_Space.norm      avgt    3     56.303 ±     1.938    B/op
 SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.churn.G1_Survivor_Space       avgt    3      0.026 ±     0.044  MB/sec
 SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.churn.G1_Survivor_Space.norm  avgt    3     ≈ 10⁻⁴                B/op
 SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.count                         avgt    3    155.000              counts
 SlsVersionMatcherBenchmark.fast_matcher_with_all_numbers:·gc.time                          avgt    3     96.000                  ms
 SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x                                       avgt    3     58.403 ±    61.797   ns/op
 SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.alloc.rate                        avgt    3  11217.666 ± 11568.783  MB/sec
+SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.alloc.rate.norm                   avgt    3    200.007 ±     0.005    B/op
 SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.churn.G1_Eden_Space               avgt    3  11271.704 ± 10899.988  MB/sec
 SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.churn.G1_Eden_Space.norm          avgt    3    200.998 ±    14.261    B/op
 SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.churn.G1_Survivor_Space           avgt    3      0.030 ±     0.058  MB/sec
 SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.churn.G1_Survivor_Space.norm      avgt    3      0.001 ±     0.001    B/op
 SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.count                             avgt    3    198.000              counts
 SlsVersionMatcherBenchmark.fast_matcher_with_one_x_x:·gc.time                              avgt    3    134.000                  ms
 SlsVersionMatcherBenchmark.matcher_with_all_numbers                                        avgt    3    243.519 ±    40.496   ns/op
 SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.alloc.rate                         avgt    3   5687.986 ±   904.582  MB/sec
+SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.alloc.rate.norm                    avgt    3    424.020 ±     0.010    B/op
 SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Eden_Space                avgt    3   5701.570 ±  1508.355  MB/sec
 SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Eden_Space.norm           avgt    3    425.023 ±    63.838    B/op
 SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Survivor_Space            avgt    3      0.014 ±     0.020  MB/sec
 SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.churn.G1_Survivor_Space.norm       avgt    3      0.001 ±     0.002    B/op
 SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.count                              avgt    3    112.000              counts
 SlsVersionMatcherBenchmark.matcher_with_all_numbers:·gc.time                               avgt    3     72.000                  ms
 SlsVersionMatcherBenchmark.matcher_with_one_x_x                                            avgt    3    252.214 ±    43.420   ns/op
 SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.alloc.rate                             avgt    3   5803.383 ±  1016.964  MB/sec
+SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.alloc.rate.norm                        avgt    3    448.021 ±     0.008    B/op
 SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Eden_Space                    avgt    3   5802.044 ±    96.422  MB/sec
 SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Eden_Space.norm               avgt    3    447.946 ±    81.868    B/op
 SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Survivor_Space                avgt    3      0.016 ±     0.085  MB/sec
 SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.churn.G1_Survivor_Space.norm           avgt    3      0.001 ±     0.006    B/op
 SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.count                                  avgt    3    114.000              counts
 SlsVersionMatcherBenchmark.matcher_with_one_x_x:·gc.time                                   avgt    3     73.000                  ms
```

iamdanfox:
Afternoon update - I tried the latest release and it's definitely improved, but hasn't got us back to a happy state.

As such, I'd quite like to squeeze all the juice we can out of this class.

@carterkozak / @CRogers are there any readability / correctness / perf changes you'd like or would you be happy to release this?

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcherParser.java
```diff
@@ -0,0 +1,181 @@
...
+                break;
+            }
+        }
+        if (next == startIndex) {
+            return fail(startIndex);
+        } else if (next == startIndex + 1) {
+            return ok(next, Character.digit(string.codePointAt(startIndex), 10));
+        } else {
+            try {
+                return ok(next, Integer.parseUnsignedInt(string.substring(startIndex, next)));
```
iamdanfox, line 122: for matches that might contain multiple digits (e.g. `10.x.x`), there is actually a new `Integer.parseUnsignedInt` method available from java9 onwards which would let me pass in this string and some indexes without needing to allocated substring.

In practise, I don't really care because the _vast_ majority of version matchers use single digit components.

schlosna (approved):

iamdanfox:
Thanks @schlosna :+1:

PR #456, opened 2021-06-08
Eliminate exceptions from SlsVersionMatcher#safeValueOf using guava's Ints.tryparse

iamdanfox:
## Before this PR

Same motivation as https://github.com/palantir/sls-version-java/pull/455

## After this PR
==COMMIT_MSG==
`SlsVersionMatcher#safeValueOf` no longer involves throwing and catching exceptions as part of its normal behaviour.
==COMMIT_MSG==

## Possible downsides?
- this is taking a guava dependency. the slsversion class gets used in quite a few APIs, so this might impose guava upgrades on consumers.

### DIFF
--- a/sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
@@ -21,6 +21,7 @@ import static com.palantir.logsafe.Preconditions.checkNotNull;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.google.common.primitives.Ints;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import java.util.Comparator;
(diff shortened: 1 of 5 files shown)


sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
```diff
@@ -165,11 +168,8 @@ public int compare(OrderableSlsVersion version) {
     }
 
     private static OptionalInt parseInt(String maybeInt) {
-        try {
-            return OptionalInt.of(Integer.parseInt(maybeInt));
-        } catch (NumberFormatException e) {
-            return OptionalInt.empty();
-        }
+        Integer maybeInteger = Ints.tryParse(maybeInt);
+        return maybeInteger != null ? OptionalInt.of(maybeInteger) : OptionalInt.empty();
```
carterkozak, line 172: double-boxing here is unfortunate, but it's a drop in the bucket.
`(int) 1` -> `Integer(1)` -> `OptionalInt(1)`

carterkozak (approved):

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
iamdanfox: guava's tryParse parses it into a boxed `Long` under the hood first and then takes it back down 😆

carterkozak: Yep, not a lot we can do without reimplementing that ourselves.

iamdanfox:
@carterkozak / @CRogers seems like autorelease didn't actually trigger here - would one of you guys be able to cut this manually for me?

PR #463, opened 2021-06-15
Hand rolled parser for SlsVersionType.RELEASE (aka x.y.z)

iamdanfox:
## Before this PR: 416 bytes / op

From 1 minute JFR, the hottest method in my application is currently `java.util.regex`, and the biggest proportion of these Matcher constructions are coming from the lovely `OrderableSlsVersion#safeValueOf`

image

This seems a bit silly, given that most of the time we're trying to parse three numbers with dots in between them.

```
Benchmark                                                         (versionString)  Mode  Cnt     Score      Error   Units
SlsVersionBenchmark.safeValueOf                                           RELEASE  avgt    3   176.778 ±   30.231   ns/op
SlsVersionBenchmark.safeValueOf:·gc.alloc.rate                            RELEASE  avgt    3  7690.289 ± 1225.982  MB/sec
SlsVersionBenchmark.safeValueOf:·gc.alloc.rate.norm                       RELEASE  avgt    3   416.021 ±    0.007    B/op
SlsVersionBenchmark.safeValueOf:·gc.churn.G1_Eden_Space                   RELEASE  avgt    3  7751.897 ± 1363.284  MB/sec
SlsVersionBenchmark.safeValueOf:·gc.churn.G1_Eden_Space.norm              RELEASE  avgt    3   419.352 ±   10.138    B/op
SlsVersionBenchmark.safeValueOf:·gc.churn.G1_Survivor_Space               RELEASE  avgt    3     0.035 ±    0.064  MB/sec
SlsVersionBenchmark.safeValueOf:·gc.churn.G1_Survivor_Space.norm          RELEASE  avgt    3     0.002 ±    0.004    B/op
SlsVersionBenchmark.safeValueOf:·gc.count                                 RELEASE  avgt    3   184.000             counts
SlsVersionBenchmark.safeValueOf:·gc.time                                  RELEASE  avgt    3   103.000                 ms
```

## After this PR: 104 bytes / op
==COMMIT_MSG==
Deserializing a verison of the form `x.y.z` now allocates less. This should mean GC algorithms have less work to do.
==COMMIT_MSG==

```

Benchmark                                                         (versionString)  Mode  Cnt     Score       Error   Units
SlsVersionBenchmark.safeValueOf                                           RELEASE  avgt    3    55.592 ±   109.978   ns/op
SlsVersionBenchmark.safeValueOf:·gc.alloc.rate                            RELEASE  avgt    3  6158.567 ± 11630.568  MB/sec
SlsVersionBenchmark.safeValueOf:·gc.alloc.rate.norm                       RELEASE  avgt    3   104.006 ±     0.002    B/op
SlsVersionBenchmark.safeValueOf:·gc.churn.G1_Eden_Space                   RELEASE  avgt    3  6200.873 ± 12743.698  MB/sec
SlsVersionBenchmark.safeValueOf:·gc.churn.G1_Eden_Space.norm              RELEASE  avgt    3   104.656 ±    22.149    B/op
SlsVersionBenchmark.safeValueOf:·gc.churn.G1_Survivor_Space               RELEASE  avgt    3     0.019 ±     0.067  MB/sec
SlsVersionBenchmark.safeValueOf:·gc.churn.G1_Survivor_Space.norm          RELEASE  avgt    3    ≈ 10⁻³                B/op
SlsVersionBenchmark.safeValueOf:·gc.count                                 RELEASE  avgt    3   147.000              counts
SlsVersionBenchmark.safeValueOf:·gc.time                                  RELEASE  avgt    3    88.000                  ms
```

### DIFF
--- a/sls-versions/src/main/java/com/palantir/sls/versions/Parsers.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/Parsers.java
@@ -1,0 +1,124 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.sls.versions;
+
+/**
+ * We're using parser combinator-style ideas here, where each parser function {@link Parsers#numberOrX},
+ * {@link Parsers#number}, {@link Parsers#literalX} accepts an index into our source string and returns a two values:
+ *
+ * - an updated index into the string representing how many characters were parsed
+ * - a value that was actually parsed out of the string (if the parser was able to parse the string), otherwise a
+ * clear signal that the parser failed (we use {@link Integer#MIN_VALUE} for this.
+ *
+ * We bit-pack these two integer values into a single long using {@link Parsers#ok} and {@link Parsers#fail} functions
+ * because primitive longs live on the stack and don't impact GC.
+ */
+final class Parsers {
+
+    static final int MAGIC_X_NUMBER = -1;
+    private static final long INT_MASK = (1L << 32) - 1;
+
+    private Parsers() {}
+
+    // "x" is signified by the magic negative number -1, which is distinct from Integer.MIN_VALUE which is a failure
+    static long numberOrX(String string, int startIndex) {
+        long xResult = literalX(string, startIndex);
+        if (isOk(xResult)) {
+            return ok(getIndex(xResult), MAGIC_X_NUMBER);
+        }
+
+        long numberResult = number(string, startIndex);
+        if (isOk(numberResult)) {
+            return numberResult;
+        }
+
+        return fail(startIndex);
+    }
+
+    static long number(String string, int startIndex) {
+        int next = startIndex;
+        int len = string.length();
+        while (next < len) {
+            int codepoint = string.codePointAt(next);
+            if (Character.isDigit(codepoint)) {
+                next += 1;
+            } else {
+                break;
+            }
+        }
+        if (next == startIndex) {
+            return fail(startIndex);
+        } else if (next == startIndex + 1) {
+            return ok(next, Character.digit(string.codePointAt(startIndex), 10));
+        } else {
+            try {
+                return ok(next, Integer.parseUnsignedInt(string.substring(startIndex, next)));
+            } catch (NumberFormatException e) {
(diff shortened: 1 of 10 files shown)


sls-versions/src/main/java/com/palantir/sls/versions/Parsers.java
```diff
@@ -0,0 +1,124 @@
...
+ * {@link Parsers#number}, {@link Parsers#literalX} accepts an index into our source string and returns a two values:
+ *
+ * - an updated index into the string representing how many characters were parsed
+ * - a value that was actually parsed out of the string (if the parser was able to parse the string), otherwise a
+ * clear signal that the parser failed (we use {@link Integer#MIN_VALUE} for this.
+ *
+ * We bit-pack these two integer values into a single long using {@link Parsers#ok} and {@link Parsers#fail} functions
+ * because primitive longs live on the stack and don't impact GC.
+ */
+final class Parsers {
```
iamdanfox, line 30: all these functions just got moved from `SlsVersionMatcherParser.java` - no changes made here.

sls-versions/src/main/java/com/palantir/sls/versions/Parsers.java
```diff
@@ -0,0 +1,124 @@
...
+    static long ok(int index, int result) {
+        return ((long) index) << 32 | (result & INT_MASK);
+    }
+
+    static long fail(int index) {
+        return ((long) index) << 32 | (Integer.MIN_VALUE & INT_MASK);
+    }
+
+    static boolean isOk(long state) {
+        return getResult(state) != Integer.MIN_VALUE;
```
CRogers: I know this is copied code, but you might as well make a `private static final int FAILED = Integer.MIN_VALUE` to make the magic number clear.

sls-versions/src/main/java/com/palantir/sls/versions/ReleaseVersionParser.java
```diff
@@ -0,0 +1,70 @@
...
+        long state = Parsers.number(string, 0);
+        if (Parsers.failed(state)) {
+            return null;
+        }
+        int first = Parsers.getResult(state);
+
+        state = Parsers.literalDot(string, Parsers.getIndex(state));
+        if (Parsers.failed(state)) {
+            return null;
+        }
```
CRogers, line 40: This is just begging for some kind of monadic bind abstraction... but obviously that would require a lot of allocation in Java.

sls-versions/src/main/java/com/palantir/sls/versions/ReleaseVersionParser.java
```diff
@@ -0,0 +1,70 @@
...
+
+        state = Parsers.number(string, Parsers.getIndex(state));
+        if (Parsers.failed(state)) {
+            return null;
+        }
+        int third = Parsers.getResult(state);
```
CRogers, line 57: I was thinking maybe you could make this less duplicated by having a for loop - but then you'd need to either make an an array or have `Int3MatchResult` be mutable/have a builder.

You could also group `literalDot`/`number` into a method, but it would need to return `Integer` so it can be nullable, which would I guess would be bad.

CRogers:
:+1:

sls-versions/src/main/java/com/palantir/sls/versions/ReleaseVersionParser.java
iamdanfox: _totally agree_

PR #466, opened 2021-06-16
[break] Stop using Matcher.group() which allocates a string

iamdanfox:
## Before this PR

For all our SlsVersionTypes:

```
    RELEASE_SNAPSHOT(Pattern.compile("^([0-9]+)\\.([0-9]+)\\.([0-9]+)-([0-9]+)-g[a-f0-9]+$"), 4),
    RELEASE(Pattern.compile("^([0-9]+)\\.([0-9]+)\\.([0-9]+)$"), 3),
    RELEASE_CANDIDATE_SNAPSHOT(Pattern.compile("^([0-9]+)\\.([0-9]+)\\.([0-9]+)-rc([0-9]+)-([0-9]+)-g[a-f0-9]+$"), 2),
    RELEASE_CANDIDATE(Pattern.compile("^([0-9]+)\\.([0-9]+)\\.([0-9]+)-rc([0-9]+)$"), 1),
    NON_ORDERABLE(Pattern.compile("^([0-9]+)\\.([0-9]+)\\.([0-9]+)(-[a-z0-9-]+)?(\\.dirty)?$"), 0);
```

We end up calling some code like:

```
OrderableSlsVersion.Builder orderableSlsVersion = new Builder()
                .type(type)
                .value(value)
                .majorVersionNumber(groups.groupAsInt(1))
                .minorVersionNumber(groups.groupAsInt(2))
                .patchVersionNumber(groups.groupAsInt(3));
```

Where each of those `groupAsInt` calls ends up allocating a string, and then immediately passing it on to the Integer.parseInt function. This enables my application to allocate a ton of strings really fast, which causes the poor GC algorithm to do work to get rid of them again. This is one of the top causes of allocation in my product.

image

## After this PR
==COMMIT_MSG==
Parsing all types of sls version no longer involves allocating strings just to parse them as integers again.
==COMMIT_MSG==

## Possible downsides?
- We have to bump from java 8 -> java 11 bytecode in order to get the `Integer.parseUnsignedInt(string, groupStart, groupEnd, RADIX);` method (which was added in java 9)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ subprojects {
     apply plugin: 'org.inferred.processors'
 
     tasks.check.dependsOn(javadoc)
-    sourceCompatibility = 1.8
+    sourceCompatibility = 11
 
     tasks.withType(JavaCompile) {
         options.compilerArgs += ['-Werror']
--- a/sls-versions/src/main/java/com/palantir/sls/versions/MatchResult.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/MatchResult.java
@@ -26,15 +26,28 @@ interface MatchResult {
     int groupCount();
 
     final class RegexMatchResult implements MatchResult {
+        private static final int RADIX = 10;
+        private final String string;
         private final Matcher matcher;
 
-        RegexMatchResult(Matcher matcher) {
+        RegexMatchResult(String string, Matcher matcher) {
+            this.string = string;
             this.matcher = matcher;
         }
 
         @Override
         public int groupAsInt(int group) {
-            return Integer.parseInt(matcher.group(group));
+            int groupStart = matcher.start(group);
+            int groupEnd = matcher.end(group);
+            if (groupStart == -1) {
+                throw new NumberFormatException();
+            }
+
+            if (groupEnd - groupStart == 1) {
+                return Character.digit(string.codePointAt(groupStart), RADIX);
+            }
+
+            return Integer.parseUnsignedInt(string, groupStart, groupEnd, RADIX);
         }
 
         @Override
--- a/sls-versions/src/main/java/com/palantir/sls/versions/MatchResult.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/MatchResult.java
@@ -43,10 +43,6 @@ interface MatchResult {
                 throw new NumberFormatException();
             }
 
-            if (groupEnd - groupStart == 1) {
-                return Character.digit(string.codePointAt(groupStart), RADIX);
-            }
-
             return Integer.parseUnsignedInt(string, groupStart, groupEnd, RADIX);
         }
 
--- a/changelog/@unreleased/pr-466.v2.yml
+++ b/changelog/@unreleased/pr-466.v2.yml
@@ -1,7 +1,6 @@
 type: break
 break:
-  description: This library no longer runs on Java 8 and requires Java 11. This allows
-    parsing of all types of sls version to no longer involve allocating strings just
-    to parse them as integers again.
+  description: This library now requires Java 11+ (previously it could run on Java
+    8+). This allows us to eliminate some String allocations from parsing SLS versions.
   links:
   - https://github.com/palantir/sls-version-java/pull/466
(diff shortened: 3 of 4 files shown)


build.gradle
```diff
@@ -44,7 +44,7 @@ subprojects {
     apply plugin: 'org.inferred.processors'
 
     tasks.check.dependsOn(javadoc)
-    sourceCompatibility = 1.8
+    sourceCompatibility = 11
```
carterkozak, line 47: We don't support 1.8 anymore, this is reasonable. Still worth classifying the PR as a `break`.

sls-versions/src/main/java/com/palantir/sls/versions/MatchResult.java
```diff
@@ -26,15 +26,28 @@
...
+            int groupEnd = matcher.end(group);
+            if (groupStart == -1) {
+                throw new NumberFormatException();
+            }
+
+            if (groupEnd - groupStart == 1) {
+                return Character.digit(string.codePointAt(groupStart), RADIX);
+            }
+
+            return Integer.parseUnsignedInt(string, groupStart, groupEnd, RADIX);
```
carterkozak: Doe the first condition actually perform better than `Integer.parseUnsignedInt`? You'd think that sort of optimization would be built in. Can't crack open the implementation on thumbs though.

iamdanfox: fair point that it's probably ok to ditch the single return Character.digit thing... it only avoids a few if statements which are probably trivially branch predicted anyway

```java
public static int parseInt(CharSequence s, int beginIndex, int endIndex, int radix)
                throws NumberFormatException {
        Objects.requireNonNull(s);

        if (beginIndex < 0 || beginIndex > endIndex || endIndex > s.length()) {
            throw new IndexOutOfBoundsException();
        }
        if (radix < Character.MIN_RADIX) {
            throw new NumberFormatException("radix " + radix +
                                            " less than Character.MIN_RADIX");
        }
        if (radix > Character.MAX_RADIX) {
            throw new NumberFormatException("radix " + radix +
                                            " greater than Character.MAX_RADIX");
        }

        boolean negative = false;
        int i = beginIndex;
        int limit = -Integer.MAX_VALUE;

        if (i < endIndex) {
            char firstChar = s.charAt(i);
            if (firstChar < '0') { // Possible leading "+" or "-"
                if (firstChar == '-') {
                    negative = true;
                    limit = Integer.MIN_VALUE;
                } else if (firstChar != '+') {
                    throw NumberFormatException.forCharSequence(s, beginIndex,
                            endIndex, i);
                }
                i++;
                if (i == endIndex) { // Cannot have lone "+" or "-"
                    throw NumberFormatException.forCharSequence(s, beginIndex,
                            endIndex, i);
                }
            }
            int multmin = limit / radix;
            int result = 0;
            while (i < endIndex) {
                // Accumulating negatively avoids surprises near MAX_VALUE
                int digit = Character.digit(s.charAt(i), radix);
                if (digit < 0 || result < multmin) {
                    throw NumberFormatException.forCharSequence(s, beginIndex,
                            endIndex, i);
                }
                result *= radix;
                if (result < limit + digit) {
                    throw NumberFormatException.forCharSequence(s, beginIndex,
                            endIndex, i);
                }
                i++;
                result -= digit;
            }
            return negative ? result : -result;
        } else {
            throw NumberFormatException.forInputString("", radix);
        }
    }
```

carterkozak:
I suppose this will be the first domino pushing our gradle plugins to require java 11 (if we haven't already merged something). Tritium and wc have already upgraded, so this shouldn't be problematic, potentially just noteworthy :-)

carterkozak (approved):

iamdanfox:
Yeah honestly I'm a bit scared to trigger the whole cascade of Java11 at build time, especially as this is going to be such a marginal gain...

Before merging - could we sanity check with other infra buddies if you guys are OK with this cascade of java11 plugins (as I imagine it'll probably trigger some increased support/user confusion)?

carterkozak:
pjf has required java 11 for some time now, but other plugins are still compatible with java 8, and we haven't made any movement toward changing that due to the lack of perceived upside.
Curious what other folks think. It might be nice to remove the legacy shaded javac dep from baseline error-prone.

robert3005:
@carterkozak @iamdanfox  our monitor definition plugin requires java 11 already and we haven't had issues

robert3005 (approved):

PR #550, opened 2022-01-06
[test-only] capture sorting behaviour of leading zeros

iamdanfox:
## Before this PR

prompted by a question from @lycarter 

## After this PR
==COMMIT_MSG==
[test-only] capture sorting behaviour of leading zeros
==COMMIT_MSG==

### DIFF
--- a/sls-versions/src/test/java/com/palantir/sls/versions/OrderableSlsVersionTests.java
+++ b/sls-versions/src/test/java/com/palantir/sls/versions/OrderableSlsVersionTests.java
@@ -28,6 +28,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.OptionalInt;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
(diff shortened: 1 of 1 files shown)


sls-versions/src/test/java/com/palantir/sls/versions/OrderableSlsVersionTests.java
```diff
@@ -228,6 +229,22 @@ void integer_overflow(String overflowVersion) {
...
+                        .map(OrderableSlsVersion::valueOf)
+                        .sorted()
+                        .map(OrderableSlsVersion::toString))
+                .describedAs("This sorting behaviour might look odd, but it's because the patch segment "
+                        + "is interpreted as an integer, and 01 and 1 are the same.")
+                .containsExactly("9.9.1", "9.9.01", "9.9.2", "9.9.02", "9.9.10");
+
+        assertThat(Stream.of("9.9.1", "9.9.01", "9.9.2", "9.9.02", "9.9.10").sorted())
+                .describedAs(
+                        "Just for comparison, here's the lexicographical ordering, " + "which is totally different")
```
carterkozak, line 244:
```suggestion
                        "Just for comparison, here's the lexicographical ordering, which is totally different")
```

carterkozak (approved):

PR #720, opened 2022-10-06
Weakly intern normalized OrderableSlsVersion instances

schlosna:
## Before this PR
Services that handle many versioned assets and artifacts spend a lot of memory and time comparing version equality. 

Note that this specific profile may be a bit biased as it does not include `DebugNonSafepoints`, see https://github.com/palantir/sls-packaging/pull/1376
image

## After this PR
==COMMIT_MSG==
Weakly intern normalized OrderableSlsVersion instances
==COMMIT_MSG==


schlosna:
closing for now

PR #763, opened 2023-02-09
Make SlsVersion serializable using Jackson ObjectWriters

pkoenig10:
## Before this PR

`SlsVersion` is not serializable. Attempting to serialize an `SlsVersion` using an `ObjectWriter` will use Jackson's default introspection based serialization and produce a value like:

```
{"value":"1.2.3","type":"RELEASE","majorVersionNumber":1,"minorVersionNumber":2,"patchVersionNumber":3}
```

Note that `SlsVersion` does serialize in the expected way when using an `ObjectMapper` because the `ObjectMapper` looks at the Jackson annotations on the subclasses. We have tests for this behavior.

## After this PR

`SlsVersion` is serializable. Attempting to serialize a `SlsVersion` using an `ObjectWriter` behaves in the expected way.

I have updated to the tests to also test serialization/deserialization with an `ObjectWriter`/`ObjectReader`.

### DIFF
--- a/sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
@@ -19,7 +19,6 @@ package com.palantir.sls.versions;
 import static com.palantir.logsafe.Preconditions.checkArgument;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.UnsafeArg;
 import java.util.Optional;
 import org.immutables.value.Value;
@@ -84,11 +83,6 @@ public abstract class OrderableSlsVersion extends SlsVersion implements Comparab
     public static boolean check(String coordinate) {
         return safeValueOf(coordinate).isPresent();
     }
-
-    @Override
-    @JsonValue
-    @Value.Auxiliary
-    public abstract String getValue();
 
     @Override
     public final String toString() {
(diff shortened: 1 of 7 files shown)


sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
```diff
@@ -85,7 +85,11 @@ public static boolean check(String coordinate) {
         return safeValueOf(coordinate).isPresent();
     }
 
+    @Override
     @JsonValue
+    @Value.Auxiliary
```
carterkozak: Would removing `@JsonValue` from the `toString` here be equivalent?

pkoenig10: I'm not sure what you're asking - this PR _does_ remove `@JsonValue` from `toString`.

carterkozak: I meant not adding a getValue override in the subclasses, since the base method could be annotated with both `@Value.Auxiliary` and `@JsonValue` unless immutables gets upset (which I suppose wouldn't be all that surprising).

pkoenig10: Sure we could do that. It just feels wrong to have immutables annotations in a non-immutables abstract class.

But happy to do whichever you prefer.

The tests also pass if I exclude the `@JsonValue` annotation on the subclasses. But that's probably not a good idea, since `@JsonValue` isn't `@Inherited` so I assume that is relying on some Jackson introspection that may or may not be guaranteed.

carterkozak (approved):

iamdanfox:
Woah - I did not expect there to be differing behaviour between jackson's ObjectMapper and ObjectWriter ... has this been affecting wire serialization of this type? if someone had persisted a value like `{"value":"1.2.3","type":"RELEASE","majorVersionNumber":1,"minorVersionNumber":2,"patchVersionNumber":3}` somewhere in a DB can it still be deserialized after your change?

carterkozak:
> can it still be deserialized after your change?


It couldn’t be deserialized before this change

PR #821, opened 2023-09-08
Do not consider versions with different hashes equal

kasper0406:
## Before this PR
TODO

## After this PR
==COMMIT_MSG==
Do not consider versions with different hashes equal
==COMMIT_MSG==

### DIFF
--- a/sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
@@ -89,7 +89,12 @@ public enum VersionComparator implements Comparator<OrderableSlsVersion> {
         // Substitute -1 if not present because snapshots are greater than non-snapshots.
         OptionalInt leftInt = left.secondSequenceVersionNumber();
         OptionalInt rightInt = right.secondSequenceVersionNumber();
-        return Integer.compare(leftInt.orElse(-1), rightInt.orElse(-1));
+        int secondSequenceCompare = Integer.compare(leftInt.orElse(-1), rightInt.orElse(-1));
+        if (secondSequenceCompare != 0) {
+            return secondSequenceCompare;
+        }
+
+        return stringCompare(left, right);
     }
 
     private int compareFirstSequenceVersions(OrderableSlsVersion left, OrderableSlsVersion right) {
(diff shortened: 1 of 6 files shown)


sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
```diff
@@ -105,6 +110,17 @@ private int compareFirstSequenceVersions(OrderableSlsVersion left, OrderableSlsV
...
+            return firstSequenceCompare;
+        }
+
+        return stringCompare(left, right);
+    }
+
+    private int stringCompare(OrderableSlsVersion left, OrderableSlsVersion right) {
+        // We know everything but the hash is equal between the two versions.
+        // We fall back to string based comparison to settle potential differences between git hashes.
+        return left.toString().compareTo(right.toString());
```
iamdanfox, line 124: one weird edge case that could bring us to this point is the case where someone has got leading zeros on their version string (e.g. "1.001.0-1-gbbb" vs "1.1.0-1-gaaa"),


but maybe that's fine to give this case an explicit order?

kasper0406: It's an interesting case. If we already accept to give an order to git hashes, I would not be too concerned just picking some arbitrary (but stable) ordering convention for this.

sls-versions/src/test/java/com/palantir/sls/versions/OrderableSlsVersionTests.java
```diff
@@ -213,9 +216,7 @@ public void testCheckWithGarbage() {
     @Test
     public void testEqualVersions() {
         assertVersionsEqual("1.0.0", "1.0.0");
-        assertVersionsEqual("1.0.0-2-gaaaaa", "1.0.0-2-gbbbbb");
```
iamdanfox, line 216: 🌶

derekjobst (approved):
This would elegantly solve a significant problem with publishing "snapshot" SLS versions from different git histories (branches) and avoiding a version collision.

iamdanfox:
The reason I'm hesitant about merging this is because as well as changing the equality, it _also_ changes the sorting behaviour of these snapshot versions.

If we decided to take this approach then I'd also need to update the `StoredOrderableVersion` class we have internally, as it doesn't currently store that extra snapshot information. This would be a DB migration (which we could do, it's just more work).

PR #942, opened 2024-04-19
start Interning instances of SlsVersionMatcher

iamdanfox:
## Before this PR

I was looking at heap histograms of apollo-catalog and noticed we're storing ~286 megabytes of this one `SlsVersionMatcher` class.

```
  12:       7159066      286362640  com.palantir.sls.versions.ImmutableSlsVersionMatcher
```

Given that I think the vast majority of SlsVersionMatchers come from a small number of constants (e.g. `1.x.x`, `2.x.x`, `3.x.x`) I figured we could eliminate pretty much _all_ of this memory usage by just interning them.

I started looking at the `@value.immutables` feature of weak interning https://immutables.github.io/immutable.html#instance-interning, but the generated code just used a ConcurrentHashMap under the hood, which made me worry about throughput. (Maybe this is because we have the `jdkOnly = true` flag on our ImmutablesStyle?)

It did however have a neat optimization for the equals method, where it just checks reference equality (i.e. `==`) and uses a separate `.isEqualTo` method which does the real equals check just for the interner.

## After this PR
==COMMIT_MSG==
start Interning instances of SlsVersionMatcher
==COMMIT_MSG==

HOWEVER I've added `do not merge` because I am a bit worried about the implications of these leading zeros in the numbers. This seems to be arising because we've marked user input string as `@Value.Auxiliary` (so it doesn't contribute to the equals check) and we're just relying on the other fields to determine equality.

```java
    @Value.Auxiliary
    @Value.Parameter
    public abstract String getValue();
```

I think we have a few options:
1. detect these leading zeros and just ban constructing objects like this
2. write a function to detect these leading zeros and run a migration to check they don't already exist inside apollo-catalog, THEN ban creating them
3. remove the `@Value.Auxiliary` annotation from the `String getValue()` and put it on all the _other_ fields, so that "01.0.0" is genuinely considered not equal to "1.0.0". I think this leads to weird implications because the `.compare` method would return 0.


pkoenig10:
I think the decision about whether the intern objects is more appropriate for the consumers of a library, than the library itself. This library has many consumers with different priorities. It's fairly easy for consumers that want to intern these objects to do so.

PR #1179, opened 2025-08-14
Use same priority for RC and RC snapshots

pkoenig10:
For a given `major.minor.patch` version, `RELEASE_CANDIDATE_SNAPSHOT` versions are not universally ordered after `RELEASE_CANDIDATE` versions. This makes the version comparison logic unnecessarily complicated.

The current meaning of `SlsVersionType.getPriority()` is not particularly well-defined. After this PR, it is more precisely defined as "a first-level ordering of versions that have the same `major.minor.patch` version".

I do not see any internal uses of `SlsVersionType.getPriority()` (and even if we did, I wouldn't expect this change to have a consequential impact).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -36,6 +36,7 @@ This specification describes _orderable_ and _non-orderable_ product version str
 Orderable version strings fall into one of 4 version types as defined by a category (release or release candidate)
 and whether it is a snapshot version or not (snapshot versions contain a commit hash at the end), the cross section
 of which produces the following:
+
  ```
 Version Type                        Example                 Format
 ------------                        -------                 ------
@@ -58,52 +59,52 @@ decisions about forward-vs-backwards product migrations. Further, it simplifies 
 via version ranges; for instance, a product may declare that it is compatible with a second product with a version in
 `[1.2.3, 2.0.0)`.
 
-For any two orderable versions, v1 and v2, we can define whether v1 is a *bigger* (equivalently, *later*, *newer*, etc)
+For any two orderable versions, `v1` and v2, we can define whether `v1` is a *larger* (equivalently, *later*, *newer*, etc)
 than v2. For the four variants, there can be up to three numeric components identifying a version. From left to right,
-they are: the usual notation of the base version (e.g., for `1.2.3`, 1=major, 2=minor, 3=patch), an optional second
+they are: the usual notation of the base version (e.g., for `1.2.3`, 1=major, 2=minor, 3=patch), an optional first
 numeric component to identify a release candidate (e.g. `-rc3`) or a snapshot version (e.g. `-5-gnm4s9ba`), and finally
 an optional third numeric component to identify a release candidate snapshot version (e.g. `-rc3-5-gnm4s9ba`).
 
-Intuitively, given the same base version, snapshot versions are bigger than non-snapshot versions, normal release
-versions are bigger than release candidate versions, and a normal release snapshot version is bigger than a release
-candidate of any kind. The following top-down procedure determines whether v1 is bigger than v2, written `v1 > v2`;
+Intuitively, the version types define an ordering of versions for a given base version:
+- For a given base version, release snapshot versions are larger than release versions
+- For a given base version, release versions are larger than release candidate and release candidate snapshot versions
+- For a given base and rc version, release candidate snapshot versions are larger than release candidate versions
+
+The following top-down procedure determines whether `v1` is larger than v2, written `v1 > v2`;
 comparisons like `major(v1) > major(v2)` are by integer ordering (not lexicographic ordering):
 
-- If `major(v1) > major(v2)`, then `v1 > v2`
-- If `minor(v1) > minor(v2)`, then `v1 > v2`
-- If `patch(v1) > patch(v2)`, then `v1 > v2`
-- From here on, let us assume that the base versions (major/minor/patch) are the same for v1 and v2
-- If v1 is a normal snapshot version and v2 is a normal release, then `v1 > v2`
-- If v1 is a normal release version and v2 is a rc version, then `v1 > v2`
-- If v1 and v2 are both normal snapshot versions and `snapshot(v1) > snapshot(v2)`, then `v1 > v2`
-- If v1 and v2 are both rc versions and `rc(v1) > rc(v2)`, then `v1 > v2`
-- From here on, let us assume that v1 and v2 are both rc versions of the same `rc()` number
-- If v1 is a snapshot rc version and v2 is a normal rc version, then `v1 > v2`
-- If v1 and v2 are both snapshot rc versions and `rcSnapshot(v1) > rcSnapshot(v2)`, then `v1 > v2`
-
-Further, v1 is as big as v2, written `v1 == v2`, iff neither `v1 > v2` nor `v2 > v1`.
-We write `v1 >= v2` if `v1 > v2` or `v1 == v2`.
+- We know nothing asbout `v1` and `v2`:
+  - If `major(v1) > major(v2)`, then `v1 > v2`
+  - If `minor(v1) > minor(v2)`, then `v1 > v2`
+  - If `patch(v1) > patch(v2)`, then `v1 > v2`
+- From here on, we know that `v1` and `v2` have the same base version:
+  - If both `v1` and `v2` are both release snapshot versions and `snapshot(v1) > snapshot(v2)`, then `v1 > v2`
+  - If `v1` is a release snapshot version and `v2` is not, then `v1 > v2`
+  - If both `v1` and `v2` are both release versions, then `v1 == v2`
+  - If `v1` is a release version and `v2` is not, then `v1 > v2`
+- From here on, we know that `v1` and `v2` have the same base version and are release candidate or release candidate snapshot version:
+  - If `rc(v1) > rc(v2)`, then `v1 > v2`
+  - If `v1` is a release candidate snapshot version and `v2` is not, then `v1 > v2`
+  - If `rcSnapshot(v1) > rcSnapshot(v2)`, then `v1 > v2`
 
 Examples, with each greater than all the previous:
-- RC: `1.0.0-rc1`
-- Bigger RC: `1.0.0-rc2`
-- RC Snapshot trumps RC: `1.0.0-rc2-4-gaaaaaaa`
-- Bigger RC Snapshot: `1.0.0-rc2-5-gccccccc`
-- Base trumps RC: `2.0.0`
-- Snapshot trumps all: `2.0.0-3-gaaaaaaa`
-- Bigger Snapshot: `2.0.0-4-gbbbbbbb`
-- Bigger Base: `2.1.0-rc1`
-- Release trumps RC: `2.1.0`
+- `1.0.0-rc1`
+- `1.0.0-rc1-1-gabcedf`
+- `1.0.0-rc1-2-gabcedf`
+- `1.0.0-rc2`
+- `1.0.0`
+- `1.0.0-1-gabcedf`
+- `1.0.0-2-gabcedf`
+- `1.0.1-rc1`
+- `1.0.1`
+- `1.1.0`
+- `2.0.0`
 
 Examples of equality:
-- `1.2.0 == 1.2.0`
-- `2.0.0-rc1 == 2.0.0-rc1`
-- `2.0.0-rc1-3-gaaaaaaa == 2.0.0-rc1-3-gbbbbbbb`
-- `2.0.0-5-gbbbbbbb == 2.0.0-5-gaaaaaaa1`
-
-Note that any two release and rc versions are equally big iff they are syntactically equal. As the second example
-demonstrates, this does not hold for snapshot versions.
-
+- `1.0.0-rc1 == 1.0.0-rc1`
+- `1.0.0-rc1-1-gaaaaaaa == 1.0.0-rc1-1-gbbbbbbb`
+- `1.0.0 == 1.0.0`
+- `1.0.0-1-gaaaaaaa == 1.0.0-1-gbbbbbbb`
 
 #### Version matchers
 
--- a/readme.md
+++ b/readme.md
@@ -59,21 +59,21 @@ decisions about forward-vs-backwards product migrations. Further, it simplifies 
 via version ranges; for instance, a product may declare that it is compatible with a second product with a version in
 `[1.2.3, 2.0.0)`.
 
-For any two orderable versions, `v1` and v2, we can define whether `v1` is a *larger* (equivalently, *later*, *newer*, etc)
-than v2. For the four variants, there can be up to three numeric components identifying a version. From left to right,
+For any two orderable versions, `v1` and `v2`, we can define whether `v1` is a *larger* (equivalently, *later*, *newer*, etc)
+than `v2`. For the four variants, there can be up to three numeric components identifying a version. From left to right,
 they are: the usual notation of the base version (e.g., for `1.2.3`, 1=major, 2=minor, 3=patch), an optional first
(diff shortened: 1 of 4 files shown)


sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
```diff
@@ -16,95 +16,54 @@
...
-        if (left.getPatchVersionNumber() != right.getPatchVersionNumber()) {
-            return left.getPatchVersionNumber() > right.getPatchVersionNumber() ? 1 : -1;
+        result = Integer.compare(
+                version1.secondSequenceVersionNumber().orElse(Integer.MIN_VALUE),
+                version2.secondSequenceVersionNumber().orElse(Integer.MIN_VALUE));
+        if (result != 0) {
+            return result;
         }
 
         return 0;
```
pkoenig10, line 67: The acute motivation for this PR is that we have a place where we need to deterministically pick the "newest" `SlsVersion`, which may include unorderable versions (where we don't particularly care which version is picked for unorderable versions as long as it is consistent).

I've debated whether this library should provide a `Comparator<SlsVersion>`. We could do this trivially by just replacing the final `return 0` with the result of the initial `getValue()` comparison.

On the one hand, it would be useful to have a consistent implementation for ordering `SlsVersion` values. On the other hand, I'm not sure want this library to endorse/support comparing values that are explicitly non-orderable.

Curious for thoughts. I think I'm leaning towards not supporting this.

sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
```diff
@@ -16,95 +16,49 @@
...
 
 /** Compares {@link OrderableSlsVersion}s by "newness", i.e., "1.4.0" is greater/newer/later than "1.2.1", etc.. */
 public enum VersionComparator implements Comparator<OrderableSlsVersion> {
     INSTANCE;
 
     @Override
-    public int compare(OrderableSlsVersion left, OrderableSlsVersion right) {
-        if (left.getValue().equals(right.getValue())) {
-            return 0;
-        }
```
pkoenig10: This comparator is now just 6 integer comparisons. This is fast enough where it's probably better if we don't do this initial full-string comparison.

schlosna: It doesn't look like there is currently a sort benchmark in https://github.com/palantir/sls-version-java/tree/develop/sls-versions/src/jmh/java/com/palantir/sls/versions but might be worth doing a quick reference equality check for `if (version1 == version2) return 0;` since `OrderableSlsVersion` are often interned in the services handling large volumes of versioned artifacts.

schlosna: An example of where this popped up previously: https://g.p.b/foundry/apollo-commons/pull/2032#discussion_r2202675

aldexis (commented):
I've got a slightly different proposal that I'd like to make. For comparison purposes, how about we consider that there are _always_ 5 components:
- Major
- Minor
- Patch
- RC
  - If absent, take it as meaning positive infinite (`Integer.MAX_VALUE`)
- Snapshot
  - If absent, take it as meaning negative infinite (`Integer.MIN_VALUE`)

Then, you can simply compare each of these in order all the time and I believe this should properly order all versions as we desire. It should also be trivial to replicate to build a deterministic ordering of unorderable sls-versions if you need to (by introducing a sixth string component that you compare after all of these).

Taking the example versions from the readme, this would give us:
| Version | Parsed |
|-|-|
| `1.0.0-rc1` | `[1, 0, 0, 1, -infinite]` |
| `1.0.0-rc1-1-gabcedf` | `[1, 0, 0, 1, 1]` |
| `1.0.0-rc1-2-gabcedf` | `[1, 0, 0, 1, 2]` |
| `1.0.0-rc2` | `[1, 0, 0, 2, -infinite]` |
| `1.0.0` | `[1, 0, 0, +infinite, -infinite]` |
| `1.0.0-1-gabcedf` | `[1, 0, 0, +infinite, 1]` |
| `1.0.0-2-gabcedf` | `[1, 0, 0, +infinite, 2]` |
| `1.0.1-rc1` | `[1, 0, 1, 1, -infinite]` |
| `1.0.1` | `[1, 0, 1, +infinite, -infinite]` |
| `1.1.0` | `[1, 1, 0, +infinite, -infinite]` |
| `2.0.0` | `[2, 0, 0, +infinite, -infinite]` |

~Side-note: I also wouldn't mind adding a util letting you get a `Comparator<SlsVersion>` by telling it _how_ you want to handle tie-breaks~ Edit: just realized that unorderable versions aren't about tie-breaking snapshots

sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
pkoenig10: Yeah I also considered this, will add it.

pkoenig10:
@aldexis I totally agree that your proposal makes the most sense and this would also be my preferred way of doing it. 

Unfortunately, the current model doesn't have "RC number" and "snapshot number" values - it has "first number" and "second number" values. As a result, the "first number" value is sometimes an RC number and sometimes a snapshot number.

| Version | `firstSequenceVersionNumber` | `secondSequenceVersionNumber` |
|-|-|-|
| `1.0.0-rc1` | 1 | None |
| `1.0.0-rc1-1-gabcedf` | 1 | 1 |
| `1.0.0` | None | None |
| `1.0.0-1-gabcedf` | 1 | None |

Are you suggesting we:
1. Redefine the meaning of `firstSequenceVersionNumber` and `secondSequenceVersionNumber`, or
2. Compute the "RC number" and "snapshot number" values in the comparison by switching over the type and use those for comparison.

If we think we can get away with the small semantic break, (1) would certainly be my preference. this would only affect `RELEASE_SNAPSHOT` versions.

Although getting this to work with the current regex-based parser implementation might be a bit tricky.

aldexis:
I think we can do this without a semantic or ABI break by:
- Adding `OptionalInt` `rcVersionNumber` and `snapshotVersionNumber` fields in OrderableSlsVersion, which pull from the corresponding field based on the version type
- Probably deprecate `firstSequenceVersionNumber` and `secondSequenceVersionNumber` (since it doesn't particularly make sense for non-orderable versions anyway)

(fwiw, I would probably have done it such that we have a single regex we use for parsing, which uses optional capturing groups, and then parse that directly into the appropriate fields, then determine the type based on the presence or not of the different fields instead, but let's start with what we have)

pkoenig10:
I'd slightly prefer not change the public API of `OrderableSlsVersion` in this PR. If we want to expose that, I'd like to handle that separately.

Even if we implement it this way, do we not think it's valuable to change the priority values? Given the current implementation that priority values are somewhat meaningless, since they don't actually define a correct ordering. Or are we worried about the semantic API break? Given that fact that the current priority values are useless for sorting, I don't imagine this would break anyone.

Even though the data model you propose makes much more sense, I'm not entirely convinced we should change the comparator implementation, primarily because it would become more expensive due to the need to perform a switch to compute the RC and snapshot versions.

aldexis:
You don't have to pay the cost of the switch/comparison every time. Just when building the immutable (or once later if you declare it as lazy). You also don't need to make the new methods part of the public API just yet. We can make these package-private for now if you think that is preferrable.

As for the priority, I wouldn't change it and would instead deprecate the field, which wouldn't be used anymore.

Here is my proposal to achieve what I'm proposing: https://github.com/palantir/sls-version-java/pull/1183 (note that I didn't change the docs, which I think likely belongs in a separate PR)

aldexis:
Per https://github.com/palantir/sls-version-java/pull/1183#issuecomment-3236452282 and https://github.com/palantir/sls-version-java/pull/1191, this approach is likely best in the end

sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
```diff
@@ -16,95 +16,54 @@
...
 /** Compares {@link OrderableSlsVersion}s by "newness", i.e., "1.4.0" is greater/newer/later than "1.2.1", etc.. */
 public enum VersionComparator implements Comparator<OrderableSlsVersion> {
     INSTANCE;
 
     @Override
-    public int compare(OrderableSlsVersion left, OrderableSlsVersion right) {
-        if (left.getValue().equals(right.getValue())) {
+    @SuppressWarnings("ImmutablesReferenceEquality")
+    public int compare(OrderableSlsVersion version1, OrderableSlsVersion version2) {
+        if (version1 == version2) {
```
aldexis, line 28: Note: This isn't _always_ beneficial on perf (in particular when comparing the same version in two distinct objects), as detailed in https://github.com/palantir/sls-version-java/pull/1191.

However, this is likely an improvement for the large majority of cases

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionType.java
```diff
@@ -23,9 +23,9 @@
  * {@link SlsVersion} objects.
  */
 public enum SlsVersionType {
-    RELEASE_SNAPSHOT(RegexParser.of("^([0-9]+)\\.([0-9]+)\\.([0-9]+)-([0-9]+)-g[a-f0-9]+$"), 4),
-    RELEASE(ReleaseVersionParser.INSTANCE, 3),
-    RELEASE_CANDIDATE_SNAPSHOT(RegexParser.of("^([0-9]+)\\.([0-9]+)\\.([0-9]+)-rc([0-9]+)-([0-9]+)-g[a-f0-9]+$"), 2),
+    RELEASE_SNAPSHOT(RegexParser.of("^([0-9]+)\\.([0-9]+)\\.([0-9]+)-([0-9]+)-g[a-f0-9]+$"), 3),
+    RELEASE(ReleaseVersionParser.INSTANCE, 2),
+    RELEASE_CANDIDATE_SNAPSHOT(RegexParser.of("^([0-9]+)\\.([0-9]+)\\.([0-9]+)-rc([0-9]+)-([0-9]+)-g[a-f0-9]+$"), 1),
```
aldexis, line 28: Note: The priority change here is unlikely to break anybody, since it is effectively impossible to leverage appropriately right now

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionType.java
```diff
@@ -45,6 +45,10 @@ Parser getParser() {
         this.priority = priority;
     }
 
+    /**
+     * Provides a first-level ordering of versions that have the same major/minor/patch versions. Version types with a
+     * higher priority are greater than version types with a lower priority.
```
aldexis, line 50: Thoughts on making it clear in the docs that there exists a case of two types sharing the same priority?
```suggestion
     * Provides a first-level ordering of versions that have the same major/minor/patch versions. Version types with a
     * higher priority are greater than version types with a lower priority.
     * Note that RELEASE_CANDIDATE and RELEASE_CANDIDATE_SNAPSHOT have the same priority.
```

readme.md
```diff
@@ -58,52 +59,52 @@ decisions about forward-vs-backwards product migrations. Further, it simplifies
 via version ranges; for instance, a product may declare that it is compatible with a second product with a version in
 `[1.2.3, 2.0.0)`.
 
-For any two orderable versions, v1 and v2, we can define whether v1 is a *bigger* (equivalently, *later*, *newer*, etc)
+For any two orderable versions, `v1` and v2, we can define whether `v1` is a *larger* (equivalently, *later*, *newer*, etc)
```
aldexis:
```suggestion
For any two orderable versions, `v1` and `v2`, we can define whether `v1` is a *larger* (equivalently, *later*, *newer*, etc)
```

readme.md
```diff
@@ -58,52 +59,52 @@ decisions about forward-vs-backwards product migrations. Further, it simplifies
...
 than v2. For the four variants, there can be up to three numeric components identifying a version. From left to right,
```
aldexis:
```suggestion
than `v2`. For the four variants, there can be up to three numeric components identifying a version. From left to right,
```

readme.md
```diff
@@ -58,52 +59,52 @@ decisions about forward-vs-backwards product migrations. Further, it simplifies
...
-they are: the usual notation of the base version (e.g., for `1.2.3`, 1=major, 2=minor, 3=patch), an optional second
+they are: the usual notation of the base version (e.g., for `1.2.3`, 1=major, 2=minor, 3=patch), an optional first
 numeric component to identify a release candidate (e.g. `-rc3`) or a snapshot version (e.g. `-5-gnm4s9ba`), and finally
 an optional third numeric component to identify a release candidate snapshot version (e.g. `-rc3-5-gnm4s9ba`).
```
aldexis:
```suggestion
an optional second numeric component to identify a release candidate snapshot version (e.g. `-rc3-5-gnm4s9ba`).
```

readme.md
```diff
@@ -58,52 +59,52 @@ decisions about forward-vs-backwards product migrations. Further, it simplifies
...
 
-Intuitively, given the same base version, snapshot versions are bigger than non-snapshot versions, normal release
-versions are bigger than release candidate versions, and a normal release snapshot version is bigger than a release
-candidate of any kind. The following top-down procedure determines whether v1 is bigger than v2, written `v1 > v2`;
+Intuitively, the version types define an ordering of versions for a given base version:
+- For a given base version, release snapshot versions are larger than release versions
+- For a given base version, release versions are larger than release candidate and release candidate snapshot versions
+- For a given base and rc version, release candidate snapshot versions are larger than release candidate versions
+
+The following top-down procedure determines whether `v1` is larger than v2, written `v1 > v2`;
```
aldexis:
```suggestion
The following top-down procedure determines whether `v1` is larger than `v2`, written `v1 > v2`;
```

readme.md
```diff
@@ -58,52 +59,52 @@ decisions about forward-vs-backwards product migrations. Further, it simplifies
...
-- If v1 is a normal release version and v2 is a rc version, then `v1 > v2`
-- If v1 and v2 are both normal snapshot versions and `snapshot(v1) > snapshot(v2)`, then `v1 > v2`
-- If v1 and v2 are both rc versions and `rc(v1) > rc(v2)`, then `v1 > v2`
-- From here on, let us assume that v1 and v2 are both rc versions of the same `rc()` number
-- If v1 is a snapshot rc version and v2 is a normal rc version, then `v1 > v2`
-- If v1 and v2 are both snapshot rc versions and `rcSnapshot(v1) > rcSnapshot(v2)`, then `v1 > v2`
-
-Further, v1 is as big as v2, written `v1 == v2`, iff neither `v1 > v2` nor `v2 > v1`.
-We write `v1 >= v2` if `v1 > v2` or `v1 == v2`.
+- We know nothing asbout `v1` and `v2`:
```
aldexis:
```suggestion
- We know nothing about `v1` and `v2`:
```

readme.md
```diff
@@ -58,52 +59,52 @@ decisions about forward-vs-backwards product migrations. Further, it simplifies
...
+  - If `major(v1) > major(v2)`, then `v1 > v2`
+  - If `minor(v1) > minor(v2)`, then `v1 > v2`
+  - If `patch(v1) > patch(v2)`, then `v1 > v2`
+- From here on, we know that `v1` and `v2` have the same base version:
```
aldexis, line 80: nitpicking/self thought: _technically_ not true, since the above doesn't talk about `v1 < v2`, but this can be reasonably assumed

aldexis (dismissed):

readme.md
pkoenig10: Yeah I spent a long time trying to find a way to structure this that balanced being comprehensive and clear. I'm open to any suggestions.

FWIW I don't think anyone is going to read this.

aldexis (approved):

PR #1183, opened 2025-08-22
Improve `OrderableSlsVersion` model and simplify comparison logic

aldexis:
## Before this PR
See https://github.com/palantir/sls-version-java/pull/1179 and discussion starting from https://github.com/palantir/sls-version-java/pull/1179#pullrequestreview-3140660469

The [comparison logic in `VersionComparator` is fairly convoluted](https://github.com/palantir/sls-version-java/blob/1799afe53d6d6febbfcd630e53a9272ecf43ecba/sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java#L30-L109), because we tried to order version types using priorities. But `RELEASE_CANDIDATE` is not universally ordered with respect to `RELEASE_CANDIDATE_SNAPSHOT` (for instance, `1.2.3-rc1 < 1.2.3-rc1-1-gabc < 1.2.3-rc2`). This means you can't simply define a priority on version types that you can leverage to order the versions of different types.

Instead, we can simply model `OrderableSlsVersion` as having 5 components:
- The already-existing major, minor and patch version numbers
- RC version number
  - When absent, it is considering as positive infinite for comparison purposes
- Snapshot version number
  - When absent, it is considering as _negative_ infinite for comparison purposes

This leads to a simple lexicographic comparison of the different numbers, in order:
| Version | Parsed |
|-|-|
| `1.0.0-rc1` | `[1, 0, 0, 1, -infinite]` |
| `1.0.0-rc1-1-gabcedf` | `[1, 0, 0, 1, 1]` |
| `1.0.0-rc1-2-gabcedf` | `[1, 0, 0, 1, 2]` |
| `1.0.0-rc2` | `[1, 0, 0, 2, -infinite]` |
| `1.0.0` | `[1, 0, 0, +infinite, -infinite]` |
| `1.0.0-1-gabcedf` | `[1, 0, 0, +infinite, 1]` |
| `1.0.0-2-gabcedf` | `[1, 0, 0, +infinite, 2]` |
| `1.0.1-rc1` | `[1, 0, 1, 1, -infinite]` |
| `1.0.1` | `[1, 0, 1, +infinite, -infinite]` |
| `1.1.0` | `[1, 1, 0, +infinite, -infinite]` |
| `2.0.0` | `[2, 0, 0, +infinite, -infinite]` |

## After this PR

==COMMIT_MSG==
Improve OrderableSlsVersion model and simplify comparison logic
==COMMIT_MSG==

### DIFF
--- a/sls-versions/src/jmh/java/com/palantir/sls/versions/SlsVersionBenchmark.java
+++ b/sls-versions/src/jmh/java/com/palantir/sls/versions/SlsVersionBenchmark.java
@@ -49,8 +49,10 @@ public class SlsVersionBenchmark {
         RELEASE("0.16.0"),
         SNAPSHOT("0.16.0-8-g116b425"),
         RC("0.16.0-rc1"),
-        RC_SNAPSHOT("0.16.0-rc1-8-g116b425.dirty"),
-        DIRTY("0.16.0-8-g116b425.dirty");
+        RC_SNAPSHOT("0.16.0-rc1-8-g116b425"),
+        DIRTY("0.16.0-8-g116b425.dirty"),
+        DIRTY_RC("0.16.0-rc1-8-g116b425.dirty"),
+        ;
 
         private final String string;
 
--- a/sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
@@ -30,7 +30,7 @@ import java.util.OptionalInt;
  *
  * <p>Bits are allocated as follows, from lowest bits to highest: <code>
  * LSB SafeLong:
- * 20 bits: distance from release
+ * 20 bits: snapshot number
  *  2 bits: priority1 (1 = RELEASE_CANDIDATE_SNAPSHOT, 0 = else; 2 and 3 are unused values)
  * 20 bits: RC number
  *  2 bits: priority2 (2 = RELEASE_SNAPSHOT, 1 = RELEASE, 0 = RELEASE_CANDIDATE; 3 is an unused value)
@@ -71,8 +71,8 @@ public final class CompactVersion implements Comparable<CompactVersion> {
     public static CompactVersion from(OrderableSlsVersion version) {
         long patch = encode20b(version.getPatchVersionNumber(), "patch");
 
-        int rcNumber = version.rcVersionNumber().orElse(0);
-        int distanceFromVersion = version.snapshotVersionNumber().orElse(0);
+        int rcNumber = version.rcNumber().orElse(0);
+        int distanceFromVersion = version.snapshotNumber().orElse(0);
 
         long lsb = encode20b(distanceFromVersion, "distanceFromVersion")
                 + (encodePriority1(version.getType()) << 20)
@@ -125,16 +125,16 @@ public final class CompactVersion implements Comparable<CompactVersion> {
         SlsVersionType type = typeFromPriority(priority1, priority2);
 
         int rcNumber = (int) (lsb >> 22) & MASK_20_BITS;
-        int distanceFromVersion = (int) lsb & MASK_20_BITS;
+        int snapshotNumber = (int) lsb & MASK_20_BITS;
 
-        OptionalInt firstSeq = OptionalInt.empty();
-        OptionalInt secondSeq = OptionalInt.empty();
+        OptionalInt rcNum = OptionalInt.empty();
+        OptionalInt snapshotNum = OptionalInt.empty();
         switch (type) {
-            case RELEASE_CANDIDATE -> firstSeq = OptionalInt.of(rcNumber);
-            case RELEASE_SNAPSHOT -> firstSeq = OptionalInt.of(distanceFromVersion);
+            case RELEASE_CANDIDATE -> rcNum = OptionalInt.of(rcNumber);
+            case RELEASE_SNAPSHOT -> snapshotNum = OptionalInt.of(snapshotNumber);
             case RELEASE_CANDIDATE_SNAPSHOT -> {
-                firstSeq = OptionalInt.of(rcNumber);
-                secondSeq = OptionalInt.of(distanceFromVersion);
+                rcNum = OptionalInt.of(rcNumber);
+                snapshotNum = OptionalInt.of(snapshotNumber);
             }
             case RELEASE, NON_ORDERABLE -> {}
         }
@@ -143,15 +143,10 @@ public final class CompactVersion implements Comparable<CompactVersion> {
                 .minorVersionNumber(minorVersionNumber)
                 .patchVersionNumber(patchVersionNumber)
                 .type(type)
-                .firstSequenceVersionNumber(firstSeq)
-                .secondSequenceVersionNumber(secondSeq)
+                .rcNumber(rcNum)
+                .snapshotNumber(snapshotNum)
                 .value(generateVersionString(
-                        majorVersionNumber,
-                        minorVersionNumber,
-                        patchVersionNumber,
-                        type,
-                        rcNumber,
-                        distanceFromVersion))
+                        majorVersionNumber, minorVersionNumber, patchVersionNumber, type, rcNumber, snapshotNumber))
                 .build();
     }
 
--- a/sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
@@ -71,14 +71,8 @@ public final class CompactVersion implements Comparable<CompactVersion> {
     public static CompactVersion from(OrderableSlsVersion version) {
         long patch = encode20b(version.getPatchVersionNumber(), "patch");
 
-        int rcNumber = version.firstSequenceVersionNumber().orElse(0);
-        int distanceFromVersion = version.secondSequenceVersionNumber().orElse(0);
-        if (version.getType().equals(SlsVersionType.RELEASE_SNAPSHOT)) {
-            // in this version format (1.0.0-10-gaaaaaa), the first sequence number represents the distance
-            // from the version rather than the implicit RC number
-            rcNumber = 0;
-            distanceFromVersion = version.firstSequenceVersionNumber().orElse(0);
-        }
+        int rcNumber = version.rcVersionNumber().orElse(0);
+        int distanceFromVersion = version.snapshotVersionNumber().orElse(0);
 
         long lsb = encode20b(distanceFromVersion, "distanceFromVersion")
                 + (encodePriority1(version.getType()) << 20)
--- a/sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
@@ -72,9 +72,9 @@ public final class CompactVersion implements Comparable<CompactVersion> {
         long patch = encode20b(version.getPatchVersionNumber(), "patch");
 
         int rcNumber = version.rcNumber().orElse(0);
-        int distanceFromVersion = version.snapshotNumber().orElse(0);
+        int snapshotNumber = version.snapshotNumber().orElse(0);
 
-        long lsb = encode20b(distanceFromVersion, "distanceFromVersion")
+        long lsb = encode20b(snapshotNumber, "snapshotNumber")
                 + (encodePriority1(version.getType()) << 20)
                 + (encode20b(rcNumber, "rcNumber") << 22)
                 + (encodePriority2(version.getType()) << 42)
@@ -146,7 +146,7 @@ public final class CompactVersion implements Comparable<CompactVersion> {
                 .rcNumber(rcNum)
                 .snapshotNumber(snapshotNum)
                 .value(generateVersionString(
-                        majorVersionNumber, minorVersionNumber, patchVersionNumber, type, rcNumber, snapshotNumber))
+                        majorVersionNumber, minorVersionNumber, patchVersionNumber, rcNum, snapshotNum))
                 .build();
     }
 
@@ -163,14 +163,14 @@ public final class CompactVersion implements Comparable<CompactVersion> {
     }
 
     private static String generateVersionString(
-            int major, int minor, int patch, SlsVersionType type, int rcNumber, int distanceFromVersion) {
+            int major, int minor, int patch, OptionalInt rcNumber, OptionalInt snapshotNumber) {
         StringBuilder sb = new StringBuilder();
         sb.append(major).append(".").append(minor).append(".").append(patch);
-        if (type.isReleaseCandidate()) {
-            sb.append("-rc").append(rcNumber);
+        if (rcNumber.isPresent()) {
+            sb.append("-rc").append(rcNumber.getAsInt());
         }
-        if (type.isSnapshot()) {
-            sb.append("-").append(distanceFromVersion).append("-gaaaaaa");
+        if (snapshotNumber.isPresent()) {
+            sb.append("-").append(snapshotNumber.getAsInt()).append("-gaaaaaa");
         }
         return sb.toString();
     }
--- a/sls-versions/src/main/java/com/palantir/sls/versions/NonOrderableSlsVersion.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/NonOrderableSlsVersion.java
@@ -21,6 +21,7 @@ import static com.palantir.logsafe.Preconditions.checkArgument;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.palantir.logsafe.UnsafeArg;
 import java.util.Optional;
+import java.util.OptionalInt;
 import org.immutables.value.Value;
 
 @Value.Immutable
@@ -50,7 +51,6 @@ public abstract class NonOrderableSlsVersion extends SlsVersion {
                 .majorVersionNumber(groups.groupAsInt(1))
                 .minorVersionNumber(groups.groupAsInt(2))
                 .patchVersionNumber(groups.groupAsInt(3))
-                .type(SlsVersionType.NON_ORDERABLE)
                 .build());
     }
 
@@ -60,6 +60,21 @@ public abstract class NonOrderableSlsVersion extends SlsVersion {
      */
     public static boolean check(String coordinate) {
         return safeValueOf(coordinate).isPresent() && !OrderableSlsVersion.check(coordinate);
+    }
+
+    @Override
+    public final OptionalInt rcNumber() {
+        return OptionalInt.empty();
+    }
+
+    @Override
+    public final OptionalInt snapshotNumber() {
+        return OptionalInt.empty();
+    }
+
+    @Override
+    public final SlsVersionType getType() {
+        return SlsVersionType.NON_ORDERABLE;
     }
 
     @Override
--- a/sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
@@ -21,7 +21,6 @@ import static com.palantir.logsafe.Preconditions.checkArgument;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.palantir.logsafe.UnsafeArg;
 import java.util.Optional;
-import java.util.OptionalInt;
 import org.immutables.value.Value;
 
 /**
@@ -31,43 +30,6 @@ import org.immutables.value.Value;
 @Value.Immutable
 @ImmutablesStyle
 public abstract class OrderableSlsVersion extends SlsVersion implements Comparable<OrderableSlsVersion> {
-
-    private static final SlsVersionType[] ORDERED_VERSION_TYPES = {
-        SlsVersionType.RELEASE,
-        SlsVersionType.RELEASE_CANDIDATE,
-        SlsVersionType.RELEASE_CANDIDATE_SNAPSHOT,
-        SlsVersionType.RELEASE_SNAPSHOT
-    };
-
-    /**
-     * The release candidate version number, if this version is a release candidate or release candidate snapshot.
-     */
-    @Value.Derived
-    OptionalInt rcVersionNumber() {
-        if (getType() == SlsVersionType.RELEASE_CANDIDATE || getType() == SlsVersionType.RELEASE_CANDIDATE_SNAPSHOT) {
-            // If the version is a release candidate (or RC snapshot),
-            //   the first sequence number is always the RC version number.
-            return firstSequenceVersionNumber();
-        }
-        return OptionalInt.empty();
-    }
-
-    /**
-     * The snapshot version number, if this version is a release snapshot or release candidate snapshot.
-     */
-    @Value.Derived
-    OptionalInt snapshotVersionNumber() {
-        if (getType() == SlsVersionType.RELEASE_SNAPSHOT) {
-            // If the version is a release snapshot, the first sequence number is always the snapshot version number
-            //   since there isn't an RC version number.
-            return firstSequenceVersionNumber();
-        } else if (getType() == SlsVersionType.RELEASE_CANDIDATE_SNAPSHOT) {
-            // For RC snapshots, the first sequence number is the RC version number,
-            //   and the second sequence number is the snapshot version number.
-            return secondSequenceVersionNumber();
-        }
-        return OptionalInt.empty();
-    }
 
     @JsonCreator
     public static OrderableSlsVersion valueOf(String value) {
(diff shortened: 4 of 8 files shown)


sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
```diff
@@ -38,6 +39,36 @@ public abstract class OrderableSlsVersion extends SlsVersion implements Comparab
         SlsVersionType.RELEASE_SNAPSHOT
     };
 
+    /**
+     * The release candidate version number, if this version is a release candidate or release candidate snapshot.
+     */
+    @Value.Derived
+    OptionalInt rcVersionNumber() {
```
aldexis: Making this and the below intentionally package-private per https://github.com/palantir/sls-version-java/pull/1179#issuecomment-3210988935

sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
```diff
@@ -16,95 +16,53 @@
...
 package com.palantir.sls.versions;
 
-import static com.palantir.logsafe.Preconditions.checkArgument;
-
-import com.palantir.logsafe.SafeArg;
 import java.util.Comparator;
-import java.util.OptionalInt;
 
 /** Compares {@link OrderableSlsVersion}s by "newness", i.e., "1.4.0" is greater/newer/later than "1.2.1", etc.. */
 public enum VersionComparator implements Comparator<OrderableSlsVersion> {
```
aldexis, line 22: Fwiw, at this point, this entire comparator could simply be:
```
Comparator.comparingInt(OrderableSlsVersion::getMajorVersionNumber)
                .thenComparingInt(OrderableSlsVersion::getMinorVersionNumber)
                .thenComparingInt(OrderableSlsVersion::getPatchVersionNumber)
                .thenComparingInt(version -> version.rcVersionNumber().orElse(Integer.MAX_VALUE))
                .thenComparingInt(version -> version.snapshotVersionNumber().orElse(Integer.MIN_VALUE));
```
but I'm not sure if this would have a performance cost, and I can't make the `VersionComparator.INSTANCE` _be_ this either

pkoenig10: For widely used library code like this, it's almost certainly better to write this ourselves instead of using a comparator with a bunch of lambdas.

sls-versions/src/main/java/com/palantir/sls/versions/CompactVersion.java
```diff
@@ -71,14 +71,8 @@ public long getLsb() {
...
-        int rcNumber = version.firstSequenceVersionNumber().orElse(0);
-        int distanceFromVersion = version.secondSequenceVersionNumber().orElse(0);
-        if (version.getType().equals(SlsVersionType.RELEASE_SNAPSHOT)) {
-            // in this version format (1.0.0-10-gaaaaaa), the first sequence number represents the distance
-            // from the version rather than the implicit RC number
-            rcNumber = 0;
-            distanceFromVersion = version.firstSequenceVersionNumber().orElse(0);
-        }
+        int rcNumber = version.rcVersionNumber().orElse(0);
+        int distanceFromVersion = version.snapshotVersionNumber().orElse(0);
```
pkoenig10:
```suggestion
        int snapshotNumber = version.snapshotVersionNumber().orElse(0);
```

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
pkoenig10: These should probably go on `SlsVersion` - that's where all other fields are declared. Even ones that are only relevant for `OrderableSlsVersion`.

pkoenig10: nit: It feels more correct to drop the version - these are just "numbers" not "version numbers". That's also consistent with verbiage we use elsewhere.

```suggestion
    /**
     * The release candidate number, if this version is a release candidate or release candidate snapshot.
     */
    @Value.Derived
    OptionalInt rcNumber() {
```

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
```diff
@@ -38,6 +39,36 @@ public abstract class OrderableSlsVersion extends SlsVersion implements Comparab
...
+            return firstSequenceVersionNumber();
+        }
+        return OptionalInt.empty();
+    }
+
+    /**
+     * The snapshot version number, if this version is a release snapshot or release candidate snapshot.
+     */
+    @Value.Derived
+    OptionalInt snapshotVersionNumber() {
```
pkoenig10: Same here

```suggestion
    /**
     * The snapshot number, if this version is a release snapshot or release candidate snapshot.
     */
    @Value.Derived
    OptionalInt snapshotNumber() {
```

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
```diff
@@ -38,6 +39,36 @@ public abstract class OrderableSlsVersion extends SlsVersion implements Comparab
...
+     * The release candidate version number, if this version is a release candidate or release candidate snapshot.
+     */
+    @Value.Derived
+    OptionalInt rcVersionNumber() {
+        if (getType() == SlsVersionType.RELEASE_CANDIDATE || getType() == SlsVersionType.RELEASE_CANDIDATE_SNAPSHOT) {
+            // If the version is a release candidate (or RC snapshot),
+            //   the first sequence number is always the RC version number.
+            return firstSequenceVersionNumber();
+        }
+        return OptionalInt.empty();
```
pkoenig10: Would prefer to use a switch without a `default` case to ensure we're considering every type. I think  it reads a bit nicer too.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
```diff
@@ -62,8 +62,20 @@ public static boolean check(String coordinate) {
 
     public abstract int getPatchVersionNumber();
 
+    /**
+     * The first version number after major, minor, and patch. Typically either RC or snapshot version number.
+     * @deprecated Use {@link OrderableSlsVersion#rcVersionNumber()} or
+     *   {@link OrderableSlsVersion#snapshotVersionNumber()} instead.
+     */
+    @Deprecated
```
pkoenig10: We probably can't deprecate these until we eliminate existing uses internally. We don't want upgrades to be blocked.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
```diff
@@ -62,8 +62,20 @@ public static boolean check(String coordinate) {
 
     public abstract int getPatchVersionNumber();
 
+    /**
+     * The first version number after major, minor, and patch. Typically either RC or snapshot version number.
```
pkoenig10: nits:
- Avoid abbrevations
- The "typically" seems unnecessary - we know exactly when this will be set.

```suggestion
     * The first version number after major, minor, and patch. The release candidate number, if this version is a release candidate or a release candidate snapshot, and the snapshot number, if this version is a release snapshot.
```

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
```diff
@@ -62,8 +62,20 @@ public static boolean check(String coordinate) {
...
+     * @deprecated Use {@link OrderableSlsVersion#rcVersionNumber()} or
+     *   {@link OrderableSlsVersion#snapshotVersionNumber()} instead.
+     */
+    @Deprecated
     public abstract OptionalInt firstSequenceVersionNumber();
 
+    /**
+     * The second version number after major, minor, and patch. Typically snapshot version number for RC snapshots.
```
pkoenig10:
```suggestion
     * The first version number after major, minor, and patch. The snapshot number, if this version is a release candidate snapshot
```

sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
```diff
@@ -16,95 +16,53 @@
...
-            return left.getMajorVersionNumber() > right.getMajorVersionNumber() ? 1 : -1;
+        result = Integer.compare(version1.getPatchVersionNumber(), version2.getPatchVersionNumber());
+        if (result != 0) {
+            return result;
         }
 
-        if (left.getMinorVersionNumber() != right.getMinorVersionNumber()) {
-            return left.getMinorVersionNumber() > right.getMinorVersionNumber() ? 1 : -1;
+        // If RC version is not present, treat it as meaning positive infinite,
+        //   as all releases/release snapshots are considered newer than any RC.
```
pkoenig10: nit: Avoid abbreviations and ambiguous `/` usage

```suggestion
        // If release candidate version is not present, treat it as the maximum value.
        // All release and release snapshots versions are newer than any release candidate or release candidate snapshot version.
```

sls-versions/src/main/java/com/palantir/sls/versions/VersionComparator.java
```diff
@@ -16,95 +16,53 @@
...
+                version1.rcVersionNumber().orElse(Integer.MAX_VALUE),
+                version2.rcVersionNumber().orElse(Integer.MAX_VALUE));
+        if (result != 0) {
+            return result;
         }
 
-        if (left.getPatchVersionNumber() != right.getPatchVersionNumber()) {
-            return left.getPatchVersionNumber() > right.getPatchVersionNumber() ? 1 : -1;
+        // If snapshot version is not present, treat it as meaning negative infinite,
+        //   as all snapshots are considered newer than the release/RC they're tied to.
```
pkoenig10:
```suggestion
        // If snapshot version is not present, treat it as the minimum value.
        // All release snapshots and release candidate snapshots versions are newer than their corresponding release and release candidate versions, respectively.
```

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
```diff
@@ -38,6 +39,36 @@ public abstract class OrderableSlsVersion extends SlsVersion implements Comparab
         SlsVersionType.RELEASE_SNAPSHOT
     };
 
+    /**
+     * The release candidate version number, if this version is a release candidate or release candidate snapshot.
+     */
+    @Value.Derived
```
pkoenig10: I think we can do better than `@Value.Derived`.

Here's a commit that:
- Makes the parsing logic a bit more explicit, allowing us to avoid `@Value.Derived`
- Avoids builder allocations
- Eliminates unnecessary fields from `ImmutableNonOrderableSlsVersion`

https://github.com/palantir/sls-version-java/commit/0069413706edad2618d982fd568ada4cf2f6157c

pkoenig10: And I'd probably take this one step further and eliminate the `firstSequenceVersionNumber` and `secondSequenceVersionNumber` fields altogether. I'm not really concerned about performance if we're not using in an internal, potentially hot, code path like like the comparator - a single, trivial, non-allocating switch statement should be more than acceptable for an accessor method.

https://github.com/palantir/sls-version-java/commit/8543c49285eca3dc5bf0ba6607b697febeb30e73

pkoenig10:
I'm on board with this change. I think I was hesitant to make a larger change to this API, but I think it's worthwhile to make the API a bit more sensible.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
aldexis: > We know exactly when this will be set.

That's not quite true, since this is a field on `SlsVersion`, not on `OrderableSlsVersion`. It's not quite determined that this _must_ be empty if not an RC or a snapshot.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
aldexis: While I agree with not wanting to block upgrades, deprecations shouldn't block upgrades (we've already had that discussion internally - if you break on deprecation warnings, this is on you to fix, since you opted into the behavior).

I'm happy to deprecate it later on, but deprecation followed by removal in a new major version seems like the appropriate path forward to me?

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
aldexis: I see in your example commit that you used `snapshotVersion` rather than `snapshotNumber`. Is there one you'd prefer? I'm even considering just `snapshot` itself fwiw

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
aldexis: While that's true, these don't make sense outside of `OrderableSlsVersion` 🤔 If anything, I'd probably want to move down/remove the `XXXsequenceVersionNumber` methods from `SlsVersion` since they're never defined in `NonOrderableSlsVersion`

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
aldexis: I like this approach (and was able to verify that we're not degrading the parsing perf/allocations either). One thing to note though is that this introduces an ABI break by removing the builder methods for `firstSequenceVersionNumber` and `secondSequenceVersionNumber` (the first one is used in at least one project internally, though usage should be trivial to fix)

sls-versions/src/jmh/java/com/palantir/sls/versions/SlsVersionBenchmark.java
```diff
@@ -49,8 +49,10 @@ public enum VersionString {
         RELEASE("0.16.0"),
         SNAPSHOT("0.16.0-8-g116b425"),
         RC("0.16.0-rc1"),
-        RC_SNAPSHOT("0.16.0-rc1-8-g116b425.dirty"),
-        DIRTY("0.16.0-8-g116b425.dirty");
+        RC_SNAPSHOT("0.16.0-rc1-8-g116b425"),
+        DIRTY("0.16.0-8-g116b425.dirty"),
+        DIRTY_RC("0.16.0-rc1-8-g116b425.dirty"),
```
aldexis, line 54: Made a change to the benchmark here, since `RC_SNAPSHOT` was actually a dirty version

sls-versions/src/main/java/com/palantir/sls/versions/NonOrderableSlsVersion.java
```diff
@@ -62,6 +62,21 @@ public static boolean check(String coordinate) {
...
+
+    @Override
+    public final OptionalInt snapshotNumber() {
+        return OptionalInt.empty();
+    }
+
+    @Override
+    public final SlsVersionType getType() {
+        return SlsVersionType.NON_ORDERABLE;
+    }
```
aldexis, line 78: As mentioned, these introduce an ABI break on the builder type. Somehow rev-api doesn't seem to run, so I'll accept these and ensure rev-api runs in a separate PR

Same comment for `OrderableSlsVersion` below

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
```diff
@@ -51,32 +44,55 @@ public static Optional<OrderableSlsVersion> safeValueOf(String value) {
             return Optional.empty();
         }
 
-        for (SlsVersionType type : ORDERED_VERSION_TYPES) {
-            MatchResult groups = type.getParser().tryParse(value);
-            if (groups != null) {
-                return Optional.of(construct(type, value, groups));
-            }
+        MatchResult groups = SlsVersionType.RELEASE.getParser().tryParse(value);
```
aldexis, line 47: I'm not particularly fond of doing multiple parsings, when we could trivially extend the `ReleaseVersionParser` to parse everything for us (and be more efficient too).

I'm going to look into that in a separate follow-up PR

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
```diff
@@ -62,9 +62,43 @@ public static boolean check(String coordinate) {
...
+    /**
+     * The release candidate version number, if this version is a release candidate or release candidate snapshot.
+     */
+    public abstract OptionalInt rcNumber();
 
-    public abstract OptionalInt secondSequenceVersionNumber();
+    /**
+     * The snapshot version number, if this version is a release snapshot or release candidate snapshot.
+     */
+    public abstract OptionalInt snapshotNumber();
```
aldexis, line 73: This is a break too, for anybody extending `SlsVersion`, though as far as I can tell, this isn't done on any of our projects

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionType.java
```diff
@@ -45,6 +45,13 @@ Parser getParser() {
         this.priority = priority;
     }
 
+    /**
+     * The priority was used to have a predetermined order of version types, to ease comparison,
+     *   but suffered problems due to RELEASE_CANDIDATE_SNAPSHOT and RELEASE_CANDIDATE versions not being systematically
+     *   ordered (e.g. 1.2.3-rc1 < 1.2.3-rc1-1-gabc < 1.2.3-rc2 < 1.2.3-rc2-1-gdef).
+     * @deprecated This field is deprecated and will be removed in a future release.
+     */
+    @Deprecated(forRemoval = true)
```
aldexis, line 54: Might need to remove this if we're still concerned about deprecations blocking upgrades

aldexis (commented):
Here are the JMH comparisons:
Develop
```
Benchmark                                           (versionString)  Mode  Cnt      Score      Error   Units
SlsVersionBenchmark.safeValueOf                             RELEASE  avgt    3     16.873 ±    0.898   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate               RELEASE  avgt    3  12659.582 ±  666.309  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm          RELEASE  avgt    3     56.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                    RELEASE  avgt    3    187.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                     RELEASE  avgt    3    106.000                 ms
SlsVersionBenchmark.safeValueOf                            SNAPSHOT  avgt    3    329.685 ±   78.529   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate              SNAPSHOT  avgt    3   9349.215 ± 2210.692  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm         SNAPSHOT  avgt    3    808.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                   SNAPSHOT  avgt    3    193.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                    SNAPSHOT  avgt    3    102.000                 ms
SlsVersionBenchmark.safeValueOf                                  RC  avgt    3    127.903 ±   15.892   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate                    RC  avgt    3  10497.571 ± 1301.039  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm               RC  avgt    3    352.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                         RC  avgt    3    179.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                          RC  avgt    3     93.000                 ms
SlsVersionBenchmark.safeValueOf                         RC_SNAPSHOT  avgt    3    276.389 ±    6.290   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate           RC_SNAPSHOT  avgt    3   8059.568 ±  184.260  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm      RC_SNAPSHOT  avgt    3    584.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                RC_SNAPSHOT  avgt    3    198.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                 RC_SNAPSHOT  avgt    3     96.000                 ms
SlsVersionBenchmark.safeValueOf                               DIRTY  avgt    3    362.697 ±    6.779   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate                 DIRTY  avgt    3   7151.281 ±  132.109  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm            DIRTY  avgt    3    680.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                      DIRTY  avgt    3    176.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                       DIRTY  avgt    3     85.000                 ms
SlsVersionBenchmark.safeValueOf                            DIRTY_RC  avgt    3    344.146 ±  336.398   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate              DIRTY_RC  avgt    3   7547.803 ± 7256.864  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm         DIRTY_RC  avgt    3    680.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                   DIRTY_RC  avgt    3    178.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                    DIRTY_RC  avgt    3     92.000                 ms
```

Current state of the PR:
```
Benchmark                                           (versionString)  Mode  Cnt      Score      Error   Units
SlsVersionBenchmark.safeValueOf                             RELEASE  avgt    3     16.798 ±    0.887   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate               RELEASE  avgt    3  12715.520 ±  667.856  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm          RELEASE  avgt    3     56.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                    RELEASE  avgt    3    189.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                     RELEASE  avgt    3    100.000                 ms
SlsVersionBenchmark.safeValueOf                            SNAPSHOT  avgt    3    270.576 ±   22.299   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate              SNAPSHOT  avgt    3  11390.574 ±  935.650  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm         SNAPSHOT  avgt    3    808.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                   SNAPSHOT  avgt    3    194.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                    SNAPSHOT  avgt    3    101.000                 ms
SlsVersionBenchmark.safeValueOf                                  RC  avgt    3    139.038 ±    3.726   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate                    RC  avgt    3   9656.714 ±  257.133  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm               RC  avgt    3    352.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                         RC  avgt    3    194.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                          RC  avgt    3     97.000                 ms
SlsVersionBenchmark.safeValueOf                         RC_SNAPSHOT  avgt    3    235.163 ±   38.654   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate           RC_SNAPSHOT  avgt    3   9472.974 ± 1551.075  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm      RC_SNAPSHOT  avgt    3    584.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                RC_SNAPSHOT  avgt    3    161.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                 RC_SNAPSHOT  avgt    3     86.000                 ms
SlsVersionBenchmark.safeValueOf                               DIRTY  avgt    3    247.986 ±   12.286   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate                 DIRTY  avgt    3  10459.128 ±  520.746  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm            DIRTY  avgt    3    680.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                      DIRTY  avgt    3    215.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                       DIRTY  avgt    3    110.000                 ms
SlsVersionBenchmark.safeValueOf                            DIRTY_RC  avgt    3    242.851 ±    1.067   ns/op
SlsVersionBenchmark.safeValueOf:gc.alloc.rate              DIRTY_RC  avgt    3  10680.313 ±   47.130  MB/sec
SlsVersionBenchmark.safeValueOf:gc.alloc.rate.norm         DIRTY_RC  avgt    3    680.000 ±    0.001    B/op
SlsVersionBenchmark.safeValueOf:gc.count                   DIRTY_RC  avgt    3    198.000             counts
SlsVersionBenchmark.safeValueOf:gc.time                    DIRTY_RC  avgt    3    101.000                 ms
```

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
pkoenig10: These fields will always be empty on a `NonOrderableSlsVersion`, right?

https://github.com/palantir/sls-version-java/blob/bc4fc923ad7b6b6b8b620d103338eddcf708db82/sls-versions/src/main/java/com/palantir/sls/versions/NonOrderableSlsVersion.java#L48-L54

aldexis: They will, but that's not part of the definition of `SlsVersion`. This is an implementation detail. I meant that _semantically_, nothing would prevent someone from building their own `NonOrderableSlsVersion` and setting these.

With that said, that isn't done as far as a I know, and I'm not sure how you'd really even do it in the first place, considering there isn't necessarily a sequence version number for non-orderable versions.

But that's why I believe these shouldn't live in `SlsVersion` in the first place and why I'd like to move `rc` and `snapshot` back down into `OrderableSlsVersion` (since they don't make sense in `NonOrderableSlsVersion`)

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
aldexis: https://github.com/palantir/sls-version-java/pull/1185

sls-versions/src/main/java/com/palantir/sls/versions/NonOrderableSlsVersion.java
aldexis: Seems like the lack of revapi check was a bug in gradle-revapi, which was fixed on develop by migrating to https://github.com/revapi/gradle-revapi as part of https://github.com/palantir/sls-version-java/pull/1184

I rebased the PR

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
pkoenig10: > But that's why I believe these shouldn't live in SlsVersion in the first place and why I'd like to move rc and snapshot back down into OrderableSlsVersion (since they don't make sense in NonOrderableSlsVersion)

Sure that makes sense. Putting these values in `SlsVersion` in the first place was probably a mistake.

sls-versions/src/main/java/com/palantir/sls/versions/SlsVersion.java
pkoenig10: > if you break on deprecation warnings, this is on you to fix, since you opted into the behavior

It would be nice if we could take this position, but I just don't think this is practical.

That being said, I don't have any problem marking these as `@Deprecated` if we do some due diligence and conclude that this won't affect very many repos and we commit to fixing them (I can't imagine there are many users reading these fields).

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
pkoenig10: I like `snapshotNumber`. It's not a version but it is a number. `snapshot` sounds to vague.

aldexis:
There are a couple of issues with this PR:
- It is unfortunately breaking backwards compatibility with Java Serializability, as described in https://github.com/palantir/sls-version-java/pull/1187#discussion_r2304150378
  - Trying to fix that is possible to some extent, but overly complex, and overall not worth the complexity
- Using another approach by defining convenience methods relying on switches actually degrades perf, for a minimal benefit (see https://github.com/palantir/sls-version-java/pull/1191 for a detailed analysis)

In the end, the best tradeoff may actually be the original https://github.com/palantir/sls-version-java/pull/1179. Ideally, we'd be able to improve the data model, but doing so has a cost on either speed or memory, so it's not particularly worth it, considering the approach in that other PR is still sensible

PR #1187, opened 2025-08-26
Explicitly set serialVersionUID and add serialization tests

pkoenig10:
The `SlsVersion` types are `Serializable` but don't explicitly set a `serialVersionUID`. This can cause serialization to break if we later add new fields or change the signature of these types.

This PR explicitly sets the `serialVersionUID` values to the current values, obtained using `ObjectStreamClass.lookup`.


sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
```diff
@@ -29,8 +30,11 @@
  */
 @Value.Immutable
 @ImmutablesStyle
+@Serial.Version(value = 1396122010215857210L)
```
pkoenig10, line 33: We may want to consider adding the `@Serial.Structural` annotation generates an explicit serialization implementation in the immutable class. This implementation uses the builder instead of setting the underlying fields directly (see the code below). This ensures that the serialization behavior is based on the semantics of the type and not implementation details of the immutable class. For example, without this it would not be backwards compatible to add a `Map` property, because the value of this field in the immutables class would be `null` instead of an empty map.

Adding `@Serial.Structural` is backwards compatible but is not forwards compatible. Existing serialized values will be deserialized directly as the immutable class, but newer values will be serialized as the inner `SerialForm`, which doesn't exist on older versions. This probably is not acceptable - we generally require persisted data to be forwards compatible, otherwise you run into issues during upgrades. You can make these kinds of canges safely by only changing the serialization logic after you are certain that all deserializers can handle the new format. But we can't do that here since adding `@Serial.Structural` generates `SerialForm` and adds the `writeReplace` method - there's no way to do the former but not the latter.

We can get away with the changes in https://github.com/palantir/sls-version-java/pull/1183 because immutables classes store optional properties as nullable fields, so this change won't break deserialization of existing serialized values. The deserialized values may violate the soft "invariants" of parsed versions (ex. a first sequence number without an RC or snapshot version), but I think that's fine - if you're using Java serialization you can't rely on new fields. Although we might be able to support this with some more complex usage of `@Value.Default`.

```java
/**
 * The serialized form captures the structural content of the value object,
 * providing the ability to reconstruct values with the capability to migrate
 * data. Uses optional, nullable, and provides flexible handling of
 * collection attributes.
 */
@Generated(from = "OrderableSlsVersion", generator = "Immutables")
private static class SerialForm implements Serializable {
  private static final long serialVersionUID = 1396122010215857210L;
  private final String[] names;
  private final Object[] values;
  SerialForm(ImmutableOrderableSlsVersion instance) {
    List<String> names = new ArrayList<>(7);
    List<Object> values = new ArrayList<>(7);
    names.add("value");
    values.add(instance.getValue());
    names.add("majorVersionNumber");
    values.add(instance.getMajorVersionNumber());
    names.add("minorVersionNumber");
    values.add(instance.getMinorVersionNumber());
    names.add("patchVersionNumber");
    values.add(instance.getPatchVersionNumber());
    if (instance.firstSequenceVersionNumber().isPresent()) {
      names.add("firstSequenceVersionNumber");
      values.add(instance.firstSequenceVersionNumber().getAsInt());
    }
    if (instance.secondSequenceVersionNumber().isPresent()) {
      names.add("secondSequenceVersionNumber");
      values.add(instance.secondSequenceVersionNumber().getAsInt());
    }
    names.add("type");
    values.add(instance.getType());
    this.names = names.toArray(new String[names.size()]);
    this.values = values.toArray();
  }

  Object readResolve() {
    OrderableSlsVersion.Builder builder = new OrderableSlsVersion.Builder();

    for (int i = 0; i < names.length; i++) {
      String name = names[i];
      if ("value".equals(name)) {
        builder.value((String) toSingle("value", values[i]));
        continue;
      }
      if ("majorVersionNumber".equals(name)) {
        builder.majorVersionNumber((Integer) toSingle("majorVersionNumber", values[i]));
        continue;
      }
      if ("minorVersionNumber".equals(name)) {
        builder.minorVersionNumber((Integer) toSingle("minorVersionNumber", values[i]));
        continue;
      }
      if ("patchVersionNumber".equals(name)) {
        builder.patchVersionNumber((Integer) toSingle("patchVersionNumber", values[i]));
        continue;
      }
      if ("firstSequenceVersionNumber".equals(name)) {
        if (values[i] != null) {
          builder.firstSequenceVersionNumber((Integer) values[i]);
        }
        continue;
      }
      if ("secondSequenceVersionNumber".equals(name)) {
        if (values[i] != null) {
          builder.secondSequenceVersionNumber((Integer) values[i]);
        }
        continue;
      }
      if ("type".equals(name)) {
        builder.type((SlsVersionType) toSingle("type", values[i]));
        continue;
      }
    }
    return builder.build();
  }

  private static Object toSingle(String attribute, Object value) {
    if (value instanceof Object[]) {
      Object[] elements = (Object[]) value;
      if (elements.length == 1) {
        return elements[0];
      }
      throw new IllegalStateException("Cannot extract scalar value for attribute '"
          + attribute + "' from array of length " + elements.length);
    }
    return value;
  }
}

private Object writeReplace() {
  return new SerialForm(this);
}
```

sls-versions/src/test/java/com/palantir/sls/versions/SlsVersionTests.java
```diff
@@ -76,4 +87,107 @@ public void testCheckWithNonOrderableVersion() {
...
+                        0,
+                        SlsVersion.valueOf("1.2.3-4-gabcdef"),
+                        new OrderableSlsVersion.Builder()
+                                .value("1.2.3-4-gabcdef")
+                                .majorVersionNumber(1)
+                                .minorVersionNumber(2)
+                                .patchVersionNumber(3)
+                                .firstSequenceVersionNumber(4)
+                                .type(SlsVersionType.RELEASE_SNAPSHOT)
+                                .build()),
```
pkoenig10, line 127: Open to suggestions for how to structure these tests. The idea is that the version number represents a specific point in time. Whenever we change the implementation in a way that affects serialization, all new test cases should use a new version number.

We will never regenerate the snapshots, but the expected values may change.

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
pkoenig10: Another option is that we "freeze" the current immutable implementation and preserve it just for use with Java serialization (by copying the immutables implementation and ripping out the stuff that is no longer necessary. We then create a new implementation not using Immutables so that we have complete control over the serialization logic.

sls-versions/src/test/java/com/palantir/sls/versions/SlsVersionTests.java
aldexis: I think it's worth a comment explaining
- What the version is about
- That we should never change the version, nor regenerate the files, but instead generate new files, so we can guarantee backwards compatibility

It might also be worth testing forward compatibility (verifying that newly serialized values can be properly deserialized with a former version), though I'm not sure how we'd do this.

sls-versions/src/test/java/com/palantir/sls/versions/SlsVersionTests.java
```diff
@@ -76,4 +86,93 @@ public void testCheckWithNonOrderableVersion() {
...
+                Arguments.of(
+                        0,
+                        SlsVersion.valueOf("1.2.3"),
+                        new OrderableSlsVersion.Builder()
+                                .value("1.2.3")
+                                .majorVersionNumber(1)
+                                .minorVersionNumber(2)
+                                .patchVersionNumber(3)
+                                .type(SlsVersionType.RELEASE)
+                                .build()),
```
aldexis, line 137: Aren't these two things equal in all cases here? I'm wondering whether we should simply use `SlsVersion.valueOf("1.2.3")` and skip the builder itself, considering this is what developers would use to generate the versions (in most cases)

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
aldexis: Unfortunately, being forward compatible limits our possibilities quite a bit. Even freezing the current implementation and adding another implementation would break it, as the new implementation would be serialized as itself (unless we implement a writeReplace logic)

This seems like a _lot_ of complexity for a problem we're not even sure is worth solving.

And another thing is, I've tried to see how rebasing https://github.com/palantir/sls-version-java/pull/1183 on top of this would work, and as expected the tests are failing, whenever there was a rc or snapshot version (or both), because both `rcNumber` and `snapshotNumber` fields are null (neither takes the value of `firstSequenceVersionNumber` as I was hoping could maybe be the case, but that makes sense).

I also can't actually write a custom `readObject` method, because that would need to live within the immutable class :(

This means I think we have only two ways to guarantee compatibility:
- "Fork" the immutable class by copying its current state, then implementing custom `readObject` and `writeObject` methods
- Keep the `XSequenceVersionNumber` fields and add `rcNumber` and `snapshotNumber` as either methods or `Value.Lazy` fields (I believe these can't be `Value.Derived` because they would otherwise not be set upon deserialization)

sls-versions/src/test/java/com/palantir/sls/versions/SlsVersionTests.java
pkoenig10: I want to be explicit about what exactly what fields are set and what fields are not.

sls-versions/src/main/java/com/palantir/sls/versions/OrderableSlsVersion.java
pkoenig10: Yeah this comment was mostly about me capturing the context from my investigation into the serialization behavior here. I agree it's not practical for us to add `@Serial.Structural` here.

sls-versions/src/test/java/com/palantir/sls/versions/SlsVersionTests.java
```diff
@@ -76,4 +86,93 @@ public void testCheckWithNonOrderableVersion() {
...
+
+    @ParameterizedTest
+    @MethodSource("serializable_arguments")
+    public void serializable(int version, SlsVersion input, SlsVersion expected) throws Exception {
+        Path path =
+                SERIALIZABLE_CASES_PATH.resolve(String.format("v" + version)).resolve(input.getValue() + ".ser");
+
+        if (!Files.exists(path)) {
+            writeObject(path, input);
+        }
```
aldexis: One nit here is that automatically generating files if they're absent is a bit surprising. It might be best to have a system property like https://github.com/palantir/conjure-java/blob/98fe01d47b838b86fb6523e6926a956d15fb2f10/conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/ParameterizedConjureGenerationTest.java#L90

However, since we don't expect to ever regenerate these, I would possibly just go ahead and remove this

aldexis (dismissed):
Let's get this merged and rebase [your PR](https://github.com/palantir/sls-version-java/pull/1179) on top of it

aldexis (approved):

PR #1190, opened 2025-08-28
Add benchmark for comparing versions

aldexis:
## Before this PR
We're looking to introduce some changes to the version comparison logic in https://github.com/palantir/sls-version-java/pull/1183 and/or https://github.com/palantir/sls-version-java/pull/1179, but we don't actually have a benchmark to determine whether these are going to impact the performance (whether positively or negatively)

## After this PR

==COMMIT_MSG==
Add benchmark for comparing versions
==COMMIT_MSG==


sls-versions/src/jmh/java/com/palantir/sls/versions/SlsVersionComparisonBenchmark.java
```diff
@@ -0,0 +1,93 @@
...
+public class SlsVersionComparisonBenchmark {
+
+    public enum Version {
+        RELEASE("0.16.0"),
+        SNAPSHOT("0.16.0-8-g116b425"),
+        RC("0.16.0-rc1"),
+        RC_SNAPSHOT("0.16.0-rc1-8-g116b425"),
+        ;
+
+        private final OrderableSlsVersion version;
```
pkoenig10:
```suggestion
        private final OrderableSlsVersion version1;
```

pkoenig10 (dismissed):

pkoenig10 (approved):

PR #1349, opened 2026-04-19
Intern SlsVersionMatcher

alongouldman:
==COMMIT_MSG==
## Before this PR
When loading releases, I saw a big memory consumption by these objects: 

```
#1  LinkedHashMap$Entry                   178M instances   10.0 GB
  #2  byte[]                                207M instances    6.8 GB
  #3  String                                207M instances    6.6 GB
  #4  HashMap$Node[]                         31M instances    4.8 GB
  #5  LinkedHashMap                          36M instances    3.2 GB
  #6  ImmutableOrderableSlsVersion           46M instances    2.6 GB 
  #7  ImmutableSlsVersionMatcher             22M instances    1.4 GB  ← version matchers
```

we can't intern `OrderableSlsVersion`, but at least we can intern `VersionMatcher`!

## After this PR
Intern VersionMatcher, so that it won't consume a lot of memory, if it's the same object.

==COMMIT_MSG==

### DIFF
--- a/sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
+++ b/sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
@@ -21,8 +21,6 @@ import static com.palantir.logsafe.Preconditions.checkNotNull;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.google.common.collect.Interner;
-import com.google.common.collect.Interners;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.logsafe.logger.SafeLogger;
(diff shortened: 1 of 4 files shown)


sls-versions/src/main/java/com/palantir/sls/versions/SlsVersionMatcher.java
```diff
@@ -48,6 +50,8 @@ public abstract class SlsVersionMatcher {
             .thenComparing(SlsVersionMatcher::getMinorVersionNumber, EMPTY_IS_GREATER)
             .thenComparing(SlsVersionMatcher::getPatchVersionNumber, EMPTY_IS_GREATER);
 
+    private static final Interner<SlsVersionMatcher> interner = Interners.newWeakInterner();
```
iamdanfox: given this is an immutables object, could you try using `@Value.Style(weakInterning = true)` ??

https://immutables.github.io/immutable.html#instance-interning

